PR #1, opened 2018-08-24
main: get rid of few panic points

lucab:
This translates a few panic points (unwrapping and unchecked indexing)
into normal error bubbling.


keszybz:
Duh, I somehow let that slip.

PR #12, opened 2020-02-06
$ZRAM_GENERATOR_ROOT, swap priority

nabijaczleweli:
This PR implements:
  * `$ZRAM_GENERATOR_ROOT` from the TODO
  * priority setting on the generated swaps – I went with `100` like the zram-tools Debian package but that's easy be easy enough to change – closing #8 
  * code cleanliness, it's more idiomatic now
  * a fix to an error output format string

PR #16, opened 2020-06-05
Rust 2018 + failure → anyhow + misc fixes

ignatenkobrain:
Based on top of #12

### DIFF
--- a/TODO
+++ b/TODO
@@ -1,8 +1,6 @@
 - instead of using the shell in the service file, consider calling the generator
   binary again and doing the initalization from it.
 - logging in generators needs to go to kmsg, println!() is not enough.
-- add a testing mode where, if $ZRAM_GENERATOR_ROOT=... is set, and do everything
-  relative to this path then.
 - figure out how to tell cargo to install into
   /usr/lib/systemd/system-generators not /usr/bin
 - make swap priority configurable?
--- a/src/main.rs
+++ b/src/main.rs
@@ -10,10 +10,11 @@ use std::borrow::Cow;
 use std::env;
 use std::fmt;
 use std::fs;
-use std::io::prelude::*;
+use std::io::{prelude::*, BufReader};
+use std::iter::FromIterator;
 use std::os::unix::fs::symlink;
 use std::path::{self, Path, PathBuf};
-use std::process::Command;
+use std::process::{Command, Stdio};
 use std::result;
 
 pub trait ResultExt<T, E>: failure::ResultExt<T, E>
@@ -32,20 +33,24 @@ impl<T, E: fmt::Display> ResultExt<T, E> for result::Result<T, E> where
     result::Result<T, E>: failure::ResultExt<T, E>
 {}
 
+fn make_parent(of: &Path) -> Result<(), Error> {
+    let parent = of.parent()
+        .ok_or_else(|| format_err!("Couldn't get parent of {}", of.display()))?;
+    fs::create_dir_all(&parent)?;
+    Ok(())
+}
+
 fn make_symlink(dst: &str, src: &Path) -> Result<(), Error> {
-    let parent = src.parent()
-        .ok_or_else(|| format_err!("Couldn't get parent of {}", src.display()))?;
-    let _ = fs::create_dir_all(&parent);
+    make_parent(src)?;
     symlink(dst, src).with_path(src)?;
     Ok(())
 }
 
 fn virtualization_container() -> Result<bool, Error> {
-    let output = match Command::new("systemd-detect-virt").arg("--container").output() {
-        Ok(ok) => ok,
-        Err(e) => return Err(format_err!("systemd-detect-virt call failed: {}", e)),
-    };
-    return Ok(output.status.success());
+    match Command::new("systemd-detect-virt").arg("--container").stdout(Stdio::null()).status() {
+        Ok(status) => Ok(status.success()),
+        Err(e) => Err(format_err!("systemd-detect-virt call failed: {}", e)),
+    }
 }
 
 fn main() {
@@ -67,7 +72,7 @@ fn main() {
         },
     };
 
-    if config.devices.len() == 0 {
+    if config.devices.is_empty() {
         println!("No devices configured, exiting.");
         std::process::exit(0);
     }
@@ -96,9 +101,7 @@ impl Device {
 
 struct Config {
     root: Cow<'static, str>,
-
     output_directory: PathBuf,
-
     devices: Vec<Device>,
 }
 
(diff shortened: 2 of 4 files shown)


src/main.rs
```diff
@@ -160,11 +173,11 @@ fn handle_device(config: &Config, device: &Device, memtotal_mb: f64) -> Result<b
...
         return Ok(false);
     }
-    
+
     let disksize = (device.zram_fraction * memtotal_mb) as u64 * 1024 * 1024;
     let service_name = format!("swap-create@{}.service", device.name);
-    println!("Creating {} for /dev/{} ({}MB)",
-             service_name, device.name, disksize / 1024 / 1024);
+    println!("Creating {} for {}dev/{} ({}MB)",
+             service_name, config.root, device.name, disksize / 1024 / 1024);
```
keszybz: I don't think prefixing /dev with the root path makes sense. We are not creating an actual device file anywhere underneath config.root.

src/main.rs
```diff
@@ -173,7 +186,7 @@ fn handle_device(config: &Config, device: &Device, memtotal_mb: f64) -> Result<b
 
     let contents = format!("\
 [Unit]
-Description=Create swap on /dev/%i
+Description=Create swap on {root}dev/%i
```
keszybz, line 180: Also here, the prefix is just misleading.

src/main.rs
```diff
@@ -182,9 +195,10 @@ DefaultDependencies=false
 [Service]
 Type=oneshot
 ExecStartPre=-modprobe zram
-ExecStart=sh -c 'echo {disksize} >/sys/block/%i/disksize'
-ExecStart=mkswap /dev/%i
+ExecStart=sh -c 'echo {disksize} >{root}sys/block/%i/disksize'
+ExecStart=mkswap {root}dev/%i
```
keszybz, line 190: No, this needs to go.

src/main.rs
```diff
@@ -235,22 +250,21 @@ fn run(config: Config) -> Result<(), Error> {
...
         modules_load.write(b"zram\n")?;
     }
 
     Ok(())
 }
 
-fn get_total_memory() -> Result<u64, Error> {
-    let path = Path::new("/proc/meminfo");
+fn get_total_memory(root: &str) -> Result<u64, Error> {
+    let path = Path::new(root).join("proc/meminfo");
```
keszybz, line 267: Here too, there is no /proc/meminfo underneath the temporary root. So the original path needs to be used.

src/main.rs
```diff
@@ -48,8 +49,17 @@ fn virtualization_container() -> Result<bool, Error> {
 }
 
 fn main() {
+    let root: Cow<'static, str> =
+        env::var("ZRAM_GENERATOR_ROOT").map(|mut root| {
+            if !root.ends_with(path::is_separator) {
+                root.push('/');
+            }
+            println!("Using {:?} as root directory", root);
+            root.into()
```
keszybz, line 52: Hmm, https://doc.rust-lang.org/std/path/struct.Path.html#method.join seems to imply that `join` deals with trailing slashes correctly. So maybe this chunk will not be necessary if the number of places below where `root` is used is reduced?

TODO
```diff
@@ -1,8 +1,6 @@
 - instead of using the shell in the service file, consider calling the generator
   binary again and doing the initalization from it.
 - logging in generators needs to go to kmsg, println!() is not enough.
-- add a testing mode where, if $ZRAM_GENERATOR_ROOT=... is set, and do everything
-  relative to this path then.
```
keszybz, line 5: There seems to have been a misunderstanding. This *does* say "everything", but actually the device paths and /proc file paths should not be touched.

src/main.rs
```diff
@@ -202,6 +202,7 @@ After={service}
 
 [Swap]
 What={root}dev/{zram_device}
+Options=pri=100
```
keszybz, line 215: Please use `Priority=100` instead.

Cargo.lock
```diff
@@ -0,0 +1,137 @@
+# This file is automatically @generated by Cargo.
+# It is not intended for manual editing.
+[[package]]
+name = "addr2line"
+version = "0.12.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a49806b9dadc843c61e7c97e72490ad7f7220ae249012fbda9ad0609457c0543"
```
keszybz: Is this really necessary? Versions of all deps are specified using the fancy version language, and this should be enough. I think the Cargo.lock approach is bonkers. This package does not get many changes, so this doesn't matter so much, but if it did, either git history would be full of pointless noise or those versions would be outdated most of the time.

src/main.rs
```diff
@@ -1,55 +1,40 @@
...
-where
-    E: fmt::Display,
-{
-    fn with_path<P: AsRef<Path>>(self, path: P) -> result::Result<T, failure::Context<String>>
-    where
-        Self: Sized,
-    {
-        self.with_context(|e| format!("{}: {}", path.as_ref().display(), e))
-    }
-}
```
keszybz, line 28: Yeah, nice to see this complex code gone...

src/main.rs
```diff
@@ -54,23 +54,14 @@ fn main() {
...
 
     if config.devices.is_empty() {
         println!("No devices configured, exiting.");
-        std::process::exit(0);
+        return Ok(());
     }
 
-    if let Err(e) = run(config) {
-        println!("{}", e);
-        std::process::exit(2);
```
keszybz, line 67: It is used to distinguish failure type in scripts. 1 == runtime failure, 2 == configuration error, etc.

src/main.rs
keszybz: I now see https://github.com/systemd/zram-generator/pull/12#issuecomment-582982875. So maybe let's leave this with Options= for now, and figure out why Priority does now work later.

Cargo.lock
nabijaczleweli: (To avoid the same discussion for the nth time here's an from years ago I got on `cargo-update` regarding not including `Cargo.toml`: https://github.com/nabijaczleweli/cargo-update/issues/50; it hinges on specifying "maj.min" instead of just "maj" dependencies, but otherwise should be relevant here too)

src/main.rs
ignatenkobrain: shall I keep it that way or you don't have much preference here?

Cargo.lock
ignatenkobrain: I was just following best practices ;) Also it is quite nice when updating some dependency to know which dependencies changed and how. But if you feel strongly against this, I can drop this part.

TODO
ignatenkobrain: I took it the way that everything should be in that root, incl meminfo so that it is easy to write unit tests because it would be matter of just generating meminfo and configuration files in some temporary directory. What should that option do then?

src/main.rs
ignatenkobrain: I think this was used only because of `{root}dev/something`, so this definitely can go now.

TODO
keszybz: After thinking about this, the approach in this patch seems OK.

src/main.rs
keszybz: I don't think this was used for anything. So if this return code stands in the way of a nice refactoring, then we can drop it.

Cargo.lock
keszybz: > what you (I?) want is the "best" version of every dependency, without having it explicitly governed by some (possibly very outdated) "ground truth". 

Exactly. I think that "best practice" is misguided. So please drop that patch for now.

ignatenkobrain: done.

src/main.rs
ignatenkobrain: I'll test this on my laptop today and check what's the problem.

PR #18, opened 2020-06-07
Call back for device set-up, template units, compression-algorithm, max-zram-size

nabijaczleweli:
This is mostly a rebase and with minor cleanups of the rewrite I did in February, so I don't really remember any prescient commentary. See individual patches for details.

My autoformatter disagreed with what had been used in #16, but I did my best to mimic the awfully verbose style by hand.


src/config.rs
```diff
@@ -0,0 +1,242 @@
...
+
+
+impl Config {
+    pub fn parse() -> Result<Config> {
+        let root: Cow<'static, str> = match env::var("ZRAM_GENERATOR_ROOT") {
+            Ok(val) => val.into(),
+            Err(env::VarError::NotPresent) => "/".into(),
+            Err(e) => return Err(e.into()),
+        };
+        println!("Using {:?} as a root directory", root);
```
nabijaczleweli, line 90: This is copied from `main.rs` in the split, so commenting here: I disagree with this stylistically – printing this on each run is hardly useful, it's better suited in the `Ok()` branch.

src/generator.rs
```diff
@@ -95,7 +94,6 @@ RemainAfterExit=yes
 ExecStartPre=-modprobe zram
 ExecStart={generator} --setup-device '%i'
 ",
```
keszybz, line 96: Apart from the executable path, this unit is now fully generic. So it could be stored as a static file in /usr/lib/systemd/system/. (Units which have a static file are in general less confusing, because people can more easily figure out where they are coming from.)

src/generator.rs
```diff
@@ -79,6 +80,8 @@ fn handle_device(output_directory: &Path, device: &Device) -> Result<bool> {
 
     let contents = format!(
         "\
+# Automatically generated by systemd-zram-generator
```
keszybz, line 82: Actually it's just `zram-generator`.

src/config.rs
```diff
@@ -127,16 +147,16 @@ impl Config {
...
-                dev.memory_limit_mb = u64::max_value();
+                dev.host_memory_limit_mb = None;
             } else {
-                dev.memory_limit_mb = val.parse().with_context(|| {
+                dev.host_memory_limit_mb = Some(val.parse().with_context(|| {
                     format!(
-                        "Failed to parse memory-limit \"{}\"",
-                        val
+                        "Failed to parse host-memory-limit \"{}\"",
+                        val,
```
keszybz: Backwards compat should be kept. So memory-limit should still be accepted.

keszybz:
Let's merge this. Minor shortcomings can be fixed later.

PR #19, opened 2020-06-08
Minor cleanups

keszybz:


src/generator.rs
```diff
@@ -128,7 +128,7 @@ After=swap-create@{zram_device}.service
 
 [Swap]
 What=/dev/{zram_device}
-Options=pri=100
+Priority=100
```
nabijaczleweli, line 131: This works for me now on systemd 245.5-3 and 245.6-1; probably fixed some time after my test on [Feb 6](https://github.com/systemd/zram-generator/pull/12#issuecomment-582982875), which would've been 244.1-2 or -3?

keszybz: I'm not aware of any changes in systemd that'd make a difference, but there might have been something, there's too many patches to remember. Either way, I think it's OK to merge, I don't think it's necessary to keep compatiblity with old systemd.

nabijaczleweli: Yeah, I've tested (although with an override on a normal swap, which shouldn't matter) on 241-7~deb10u4 (up-to-date buster) and it worked fine there too. Dunno why it didn't work before.

Issue #14, opened 2020-02-07
Create devices if not present?


nabijaczleweli:
Similarly to what was causing hangs in https://github.com/systemd/zram-generator/commit/5195ea0e1f27361275db123320c3b6f1e1c32641 – by default the `zram` module creates one device (as per [sexion 1 "Load module"](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/admin-guide/blockdev/zram.rst?id=5871023c3a316e8fe8c03e7bad1d35057e69ab40)), but `swap.target` will hang indefinitely if it depends on `dev-zram1.swap`, for example.

It should be simple enough to `modprobe` and read `/sys/class/zram-control/hot_add` (sexion 7 "Add/remove zram devices") if the requisite devices are not present [here](https://github.com/nabijaczleweli/systemd-zram-generator/blob/8c83cc366b00cfd52fba3ff05c89c35fb8f00322/src/generator.rs#L49) until they become so, but that's a design decision, which I'll defer to the maintainer before implementing.

keszybz:
I'm not sure. Would the algorithm be to read `hot_add` until we get a number great or equal to the desired device number? I guess that's an option.

nabijaczleweli:
In essence, yes; the main problem with that is that we don't have a *time* for it.

On the generation run we might not have the module loaded yet, and when it *is* loaded we might not be invoked again because we're waiting for the setup run on, e.g. /dev/zram4 which simply isn't going to happen.

A glance at `modules-load.d(5)` doesn't reveal any way to specify any arguments for the module (like `num-devices=` for `zram`), either.

The only solution I came up with so far would be to ditch sd-modules-load and exec `modprobe zram` during the generation run instead in the `if devices_made` branch, then check if all devices we want are there and poke `hot_add` as necessary.

nabijaczleweli:
Proposed impl in https://github.com/nabijaczleweli/systemd-zram-generator/commit/4e4859bc96db108550349b131bf1f34025420929.

PR #20, opened 2020-06-08
Split out swap-create@.service into a file to be installed separately

nabijaczleweli:


swap-create@.service
```diff
@@ -0,0 +1,14 @@
...
+Description=Create swap on /dev/%i
+Wants=systemd-modules-load.service
+After=systemd-modules-load.service
+After=dev-%i.device
+DefaultDependencies=false
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStartPre=-modprobe zram
```
nabijaczleweli, line 13: I'm not sure if this is necessary, seeing as we schedule ourselves after not only `systemd-modules-load.service`, but `dev-%i.device` as well, which can only be provided by the zram module.

nabijaczleweli:
The `dev-{}.swap` units are only parameterised on that one substituted part. Is there a way to make those a template as well?

nabijaczleweli:
Rebased.

swap-create@.service
keszybz: Hmm, indeed. Let's keep this for now, it doesn't hurt.

swap-create@.service
```diff
@@ -0,0 +1,14 @@
+# SPDX-License-Identifier: MIT
```
keszybz, line 1: Maybe create a `units/` subdirectory to store files? I seems likely we'll have more at some point.

swap-create@.service
```diff
@@ -0,0 +1,14 @@
...
+Wants=systemd-modules-load.service
+After=systemd-modules-load.service
+After=dev-%i.device
+DefaultDependencies=false
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStartPre=-modprobe zram
+ExecStart={generator} --setup-device '%i'
```
keszybz, line 14: I think the template should just say `/usr/lib/systemd/system-generators/zram-generator`. That's the only place where the generator is supposed to go (as long as split-usr systems are ignored and a few other details ;)).

src/generator.rs
```diff
@@ -76,40 +73,6 @@ pub fn run_generator(
...
-    fs::write(&service_path, contents).with_context(|| {
-        format!(
-            "Failed to write a device service into {}",
-            service_path.display()
-        )
-    })?;
-
-    Ok(())
-}
-
```
keszybz, line 112: Hmm, maybe include a drop-in that overrides ExecStart if the generator is run from a non-standard location? This will make testing a lot easier.

README.md
```diff
@@ -22,6 +22,12 @@ This program serves two purposes:
 2. Once we figure out all the details, it should be useful as an
    example of a systemd generator in rust.
 
+### Installation
+
+The "{generator}" template `swap-create@.service` must be substituted
+for the final location of the zram-generator binary and copied into
+a system unit directory.
```
keszybz, line 29: Doesn't cargo have a way to install data files? Telling users to put files in various locations is very error prone.

nabijaczleweli: It doesn't, owing to the fact that cargo is just a libdep resolver (and you can build an executable if you absolutely *must*), and not a distribution/package manager.

I usually just use makefiles.

keszybz: OK. So let's just hardcode the value in the template.

swap-create@.service
qhill: Why not reference modprobe@zram.service instead?

README.md
```diff
@@ -22,6 +22,14 @@ This program serves two purposes:
 2. Once we figure out all the details, it should be useful as an
    example of a systemd generator in rust.
 
+### Installation
+
+Copy the `zram-generator` binary into `/usr/lib/systemd/zram-generators/`,
```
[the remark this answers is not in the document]
nabijaczleweli: Fixed

README.md
```diff
@@ -22,6 +22,14 @@ This program serves two purposes:
...
+and `units/swap-create@.service` into `/usr/lib/systemd/system/`.
+
+The "{generator}" template in `units/swap-create@.service.d/binary-location.conf`
```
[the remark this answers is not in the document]
nabijaczleweli: Well, we're already defaulting to /usr for the generator, IMO it's easier to understand that you can copy it to /etc or /run or whatever if we say to copy *both* to /usr instead of distributing them over /usr and /etc

swap-create@.service
nabijaczleweli: Because leaving it as-is is cleaner than changing it one way now, and the other in the next patch I'm gonna send (cf. #14).

PR #21, opened 2020-06-09
Tests

keszybz:


Cargo.toml
```diff
@@ -10,6 +10,8 @@ edition = "2018"
 [dependencies]
 anyhow = "1"
 rust-ini = "0.13"
+structopt = "0.3"
+tempfile = "3"
```
ignatenkobrain, line 14: this should be in `dev-dependencies` section

src/lib.rs
```diff
@@ -0,0 +1,5 @@
+/* SPDX-License-Identifier: MIT */
+
+pub mod config;
+pub mod generator;
+pub mod setup;
```
ignatenkobrain, line 5: hmm... do we really need to make a library out of this?

src/main.rs
```diff
@@ -4,9 +4,55 @@ mod config;
...
+
+    arg: String,
+    extra: Vec<String>
+}
+
+fn get_opts() -> Result<Opts> {
+    let opts = Opts::from_args();
+    println!("{:?}", opts);
+
+    if opts.setup_device && opts.extra.len() != 0 {
```
ignatenkobrain, line 29: `opts.extra.is_empty()`

tests/test_cases.rs
```diff
@@ -0,0 +1,88 @@
+use zram_generator::{config, generator};
+
+use anyhow::Error;
+use tempfile::TempDir;
+use std::fs;
+use std::path::Path;
+use std::process::Command;
+
+fn prepare_directory(srcroot: &Path) -> Result<TempDir, Error> {
```
ignatenkobrain, line 9: I think easier to use Result from anyhow, so it would look just like `-> Result<TempDir>`

src/config.rs
```diff
@@ -249,3 +178,43 @@ fn get_total_memory_kb(root: &str) -> Result<u64> {
...
 }
+
+fn get_total_memory_kb(root: &Path) -> Result<u64> {
+    let path = root.join("proc/meminfo");
+    _get_total_memory_kb(&path)
+}
+
+#[cfg(test)]
+mod tests {
+    extern crate tempfile;
```
ignatenkobrain, line 189: no need to use `extern crate` since zram-generator is Rust 2018

src/lib.rs
keszybz: Yeah, I was wondering that too. Do you have a better way to allow `config` and `generator` to be imported?

keszybz:
Updated with the suggested changes.

src/lib.rs
ignatenkobrain: Apparently there is no better way when having tests in tests/ directory that would like to use library instead of calling to a binary. In future we might want to move the unit-tests to the src/ and have tests/ to just call a binary interface but I think it is good enough for now. Any testing is better than no testing :)

PR #24, opened 2020-06-12
Create simple makefile to simplify installation procedure

Prototik:
Contributions welcome

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,21 @@
+INSTALL_PROGRAM = install
+CARGO_PROGRAM = cargo
+PREFIX = /usr
+SYSTEMD_DIR ?= $(PREFIX)/lib/systemd
+
+.DEFAULT: build
+.PHONY: build check clean install
+
+build:
+	@$(CARGO_PROGRAM) build --release
+
+check: build
+	@$(CARGO_PROGRAM) test --release
+
+clean:
+	@$(CARGO_PROGRAM) clean
+
+install: build
+	$(INSTALL_PROGRAM) -Dm755 target/release/zram-generator $(DESTDIR)$(SYSTEMD_DIR)/system-generators/zram-generator
+	$(INSTALL_PROGRAM) -Dm644 units/swap-create@.service $(DESTDIR)$(SYSTEMD_DIR)/system/swap-create@.service
+	$(INSTALL_PROGRAM) -Dm644 zram-generator.conf.example $(DESTDIR)$(PREFIX)/share/doc/zram-generator/zram-generator.conf.example
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
-INSTALL_PROGRAM = install
-CARGO_PROGRAM = cargo
+INSTALL = install
+CARGO = cargo
 PREFIX = /usr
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -0,0 +1,21 @@
+INSTALL_PROGRAM = install
```
ignatenkobrain, line 1: I think this should be just `INSTALL ?= install`

Makefile
```diff
@@ -0,0 +1,21 @@
...
+CARGO_PROGRAM = cargo
```
ignatenkobrain, line 2: same here, just a `CARGO`

Makefile
```diff
@@ -0,0 +1,21 @@
...
+check: build
+	@$(CARGO_PROGRAM) test --release
+
+clean:
+	@$(CARGO_PROGRAM) clean
+
+install: build
+	$(INSTALL_PROGRAM) -Dm755 target/release/zram-generator $(DESTDIR)$(SYSTEMD_DIR)/system-generators/zram-generator
+	$(INSTALL_PROGRAM) -Dm644 units/swap-create@.service $(DESTDIR)$(SYSTEMD_DIR)/system/swap-create@.service
+	$(INSTALL_PROGRAM) -Dm644 zram-generator.conf.example $(DESTDIR)$(PREFIX)/share/doc/zram-generator/zram-generator.conf.example
```
ignatenkobrain, line 21: these should have `-p` to preserve timestamps.

README.md
```diff
@@ -24,9 +24,11 @@ This program serves two purposes:
...
-Copy the `zram-generator` binary  into `/usr/lib/systemd/system-generators/`,
-    `units/swap-create@.service`  into `/usr/lib/systemd/system/`,
-and `zram-generator.conf.example` into `/etc/systemd/zram-generator.conf`, customising it to your liking.
+Executing `make install` will create the following things:
+* Generator binary installed as `/usr/lib/systemd/system-generators/zram-generator`
+* `units/swap-create@.service` copied into `/usr/lib/systemd/system/`
+* `zram-generator.conf.example` copied into `/usr/share/doc/zram-generator/`
+You need though create your own config file at `/etc/systemd/zram-generator.conf`, customising it to your liking.
 
 The "{generator}" template in `units/swap-create@.service.d/binary-location.conf`
```
nabijaczleweli, line 33: This is better moved into the Testing sexion now.

Prototik:
> How about an `uninstall` target that removes the three files?

That's not common usecase, usually `install` target used in packaging scripts and uninstalling handled by package manager, not makefiles.

nabijaczleweli:
I've seen it around, but more than fair enough.

Makefile
```diff
@@ -0,0 +1,21 @@
+INSTALL_PROGRAM = install
+CARGO_PROGRAM = cargo
+PREFIX = /usr
+SYSTEMD_DIR ?= $(PREFIX)/lib/systemd
```
keszybz, line 4: This could also be `pkgconf systemd --variable=systemd_system_unit_dir`, but that is fine too.

keszybz:
```
/usr/share/rust/.cargo/bin/cargo test --verbose
error: failed to download `remove_dir_all v0.5.2`
```
cargo seems to be having a bad day.

PR #25, opened 2020-06-12
Create zram devices if not present

### LINKED ISSUE #14: Create devices if not present?

### LINKED ISSUE #26: generators should not write out anything but units
The generator writes out a drop-in for load-modules.d. Not OK, will break during daemon-reload.

As documented, please only generate unit files, nothing else.

https://github.com/systemd/zram-generator/blob/943d7aaf47a19c973310616cbd28b0e8f7455146/src/generator.rs#L53

nabijaczleweli:
Closes #14
Closes #26


nabijaczleweli:
CI failure appears unrelated.

src/generator.rs
```diff
@@ -44,26 +46,36 @@ pub fn run_generator(root: &Path, devices: &[Device], output_directory: &PathBuf
...
+        let max_device = devices_made.into_iter().fold(0, cmp::max);
+        if !Path::new("/dev")
+            .join(format!("zram{}", max_device))
+            .exists()
+        {
+            while fs::read_to_string("/sys/class/zram-control/hot_add")
+                .context("Adding zram device")?
+                .trim_end()
+                .parse::<u64>()
+                .context("Fresh zram device number")?
```
keszybz, line 69: I don't want this to happen in the "generation" phase. Loading of modules is a potentially lengthy operation (because the kernel might need to pull in dependencies first, read stuff off the filesystem, etc). Generators are supposed to run quickly and without blocking.

Instead, this should be moved to the setup phase.

units/swap-create@.service
```diff
@@ -2,13 +2,10 @@
 
 [Unit]
 Description=Create swap on /dev/%i
-Wants=systemd-modules-load.service
-After=systemd-modules-load.service
```
keszybz, line 6: Dropping `ExecStartPre=-modprobe` is OK. But please keep this part. We generally still prefer to load modules via `systemd-modules-load.service` then ourselves.

tests/test_cases.rs
```diff
@@ -76,11 +76,13 @@ fn test_generation(name: &str) -> Result<Vec<config::Device>> {
 }
 
 #[test]
+#[ignore] // Needs root for the modprobe call
 fn test_01_basic() {
     test_generation("01-basic").unwrap();
 }
 
 #[test]
+#[ignore] // Needs root for the modprobe call
```
keszybz, line 85: Instead of requiring manual manipulation of which tests are run, it'd be much better to simply exit if we lack privileges. With a growing number of tests, it'd soon become very cumbersome to enable and disable tests by hand like this.

src/generator.rs
nabijaczleweli: As noted in https://github.com/systemd/zram-generator/issues/14#issuecomment-640725858, the setup phase will simply *never run* if we don't do this, just like it doesn't now for zram devs higher than the default num_devices (usually 1).

tests/test_cases.rs
nabijaczleweli: There's no way to mark a test skipped from inside said test, so the test can either fail (which, well, isn't really true) or succeed (which is an even worse "the tests succeed but don't actually test anything and are entirely useless" false positive). This is the only way to ensure vaguely correct semantics.

units/swap-create@.service
nabijaczleweli: https://github.com/systemd/zram-generator/issues/26: `modules-load.d` cannot be used, so ordering oneself after it seems foolish; marking resolved.

tests/test_cases.rs
nabijaczleweli: Updated to only run modprobe if the module doesn't exist, so doesn't require root.

nabijaczleweli:
Rebased; updated: the modprobe call is only done if needed.

The only alternative for this is generating *another* stage, with the sole purpose of execing `modprobe zram` and looping `read("sys/class/zram-control/hot_add")`. This feels like more overhead and complexity than is worth for me. The modprobe call is gonna be paid exactly *once* and it'll have to be done, anyway.

nabijaczleweli:
Rebased

poettering:
Hmm, do I get this right, this now busy loops after hot_add until the device popped up? That's terrible!

I mean, the zram module appears quite broken to expose such a borked API, and that should be fixed. If you want to work around that I'd propose however:

1. ship some service that does `modprobe -qab zram` and is ordered before your .swap unit
2. ship a static drop-in for /usr/lib/modprobe.d/ (with some early name) that sets the module's num_devices to 1 (it is zero by default, iiuc?).

keszybz:
> Hmm, do I get this right, this now busy loops after hot_add until the device popped up? That's terrible!
> 
> I mean, the zram module appears quite broken to expose such a borked API, and that should be fixed. If you want to work around that I'd propose however:
> 
>     1. ship some service that does `modprobe -qab zram` and is ordered before your .swap unit
> 
>     2. ship a static drop-in for /usr/lib/modprobe.d/ (with some early name) that sets the module's num_devices to 1 (it is zero by default, iiuc?).

The default is 1. So normally, after loading of the module we will not do anything, because the device already exists. This `hot_add` path is only a fallback for when the user declares a different device number than zram0. (It is possible, and useful e.g. to have a compressed cache partition in ram, but probably very very rarely used.)

I agree that this is fairly ugly, but meh, it shouldn't matter that much in practice.

PR #31, opened 2020-06-15
Reset devices and raise limits

keszybz:


src/generator.rs
```diff
@@ -24,6 +24,7 @@ fn make_symlink(dst: &str, src: &Path) -> Result<()> {
 
 fn virtualization_container() -> Result<bool> {
     match Command::new("systemd-detect-virt")
+        .arg("--quiet")
```
nabijaczleweli, line 27: This probably means the stdout redirection below can be dropped

nabijaczleweli:
The option parsing has turned into a mess of bools enforcing invariants that clap (which structopt already uses as its back-end) already has the capability to enforce. Feel free to merge in (should be FF) this commit: https://github.com/nabijaczleweli/systemd-zram-generator/commit/1c3137a5ca8f1f22c0fd755d4b842a61f2dae0b4, which fixes that by using clap directly (and also fixes formatting).

keszybz:
Thanks, updated.

Cargo.toml
```diff
@@ -3,14 +3,14 @@ name = "zram-generator"
...
 authors = ["Zbigniew Jędrzejewski-Szmek <zbyszek@in.waw.pl>"]
 license = "MIT"
-description = "Systemd unit generator for zram devices"
+description = "Systemd unit generator for zram swap devices"
 homepage = "https://github.com/systemd/zram-generator"
 edition = "2018"
 
 [dependencies]
-anyhow = "1"
+anyhow = "1.0"
```
ignatenkobrain, line 11: hm, what is the reason for this change?

keszybz: No idea. But it shouldn't be part of this patch even if it is otherwise justified, so I dropped it.

Cargo.toml
```diff
@@ -3,14 +3,14 @@ name = "zram-generator"
...
 authors = ["Zbigniew Jędrzejewski-Szmek <zbyszek@in.waw.pl>"]
 license = "MIT"
-description = "Systemd unit generator for zram devices"
+description = "Systemd unit generator for zram swap devices"
 homepage = "https://github.com/systemd/zram-generator"
 edition = "2018"
 
 [dependencies]
 anyhow = "1"
+clap = "2.33"
```
ignatenkobrain, line 12: One useful thing to do here is to change it to:

`clap = { version = "2.33", default-features = false }`

the coloring and some guessing of subcommands that we don't use is not really needed, but simplifies dependency chain quite a bit.

```diff
--- before	2020-06-16 07:28:46.253611116 +0200
+++ after	2020-06-16 07:28:57.112737416 +0200
@@ -1,13 +1,8 @@
 zram-generator v0.1.2 (/home/brain/Projects/upstream/zram-generator)
 ├── anyhow v1.0.31
 ├── clap v2.33.1
-│   ├── ansi_term v0.11.0
-│   ├── atty v0.2.14
-│   │   └── libc v0.2.71
 │   ├── bitflags v1.2.1
-│   ├── strsim v0.8.0
 │   ├── textwrap v0.11.0
 │   │   └── unicode-width v0.1.7
-│   ├── unicode-width v0.1.7
-│   └── vec_map v0.8.2
+│   └── unicode-width v0.1.7
 └── rust-ini v0.13.0
```

ignatenkobrain:
Apart from one small comment, looks good.

Cargo.toml
nabijaczleweli: Right, I had this kicking about in my Cargo.toml and missed it; soz.

Maryse47:
@keszybz you missed updating of example config: https://github.com/systemd/zram-generator/blob/master/zram-generator.conf.example

PR #32, opened 2020-06-16
[ex-RFC] Use kernlog(/+simplelog) for logging

nabijaczleweli:
There's two ways of tying stdout/err to `/dev/kmsg` I've found so far:
1. [`gag`](https://docs.rs/gag/0.1.10/gag), which `dup2()`s an fd onto 1/2. The problem with that is /dev/kmsg's peculiar format, which'd have to be reproduced on each `println!()` call.
2. [`kernlog`](https://docs.rs/kernlog/0.2.1/kernlog/), which hooks onto the de-facto-standard `log` crate and manages that behind the scenes.

`kernlog` will, obviously, fail to initialise if it can't write to `/dev/kmsg` (usually this means it isn't root). This can be worked around in a few ways:
1. The maxi approach taken here, using another logger (here: `simplelog`, which is rather large dependency-wise), if kernlog doesn't initialise.
2. The midi approach, using, for example, a "develop" feature, which will enable `simplelog` only for builds you specify (I think one could be devilish and enable it only on the debug build profile, but).
3. The mini approach: simply not. If the logger initialises, great; if not: whatever. If no `log` logger is installed, the output is ignored.
4. The micro approach: `kernlog::init().unwrap()` or equivalent.

I've PRed `kernlog.rs` with an update from the 2015 state of affairs (https://github.com/kstep/kernlog.rs/pull/2), to no response so far; hence the `git=` dep.

screenshot of two "zram-generator" invocations; first prefixed by sudo with output showing up in the journal, second without, with output showing up in the terminal

The `::module` thingies can be overriden with `warn!(target: "systemd-zram", ...)`, if desired

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,9 +14,8 @@ anyhow = "1.0.12"
 clap = { version = "2.33", default-features = false }
 liboverdrop = "0.0.2"
 rust-ini = ">=0.13, <0.16"
-kernlog = { git = "https://github.com/nabijaczleweli/kernlog.rs.git", rev = "a44a3f021815e091966741aa3bcdcefdfd51590d" }
-simplelog = { version = "0.8", default-features = false }
-log = "0.4"
+libc = "0.2"
+log = { version = "0.4", features = ["std"] }
 
 [dev-dependencies]
 tempfile = "3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,9 @@ anyhow = "1.0.12"
 clap = { version = "2.33", default-features = false }
 liboverdrop = "0.0.2"
 rust-ini = ">=0.13, <0.16"
+kernlog = { git = "https://github.com/nabijaczleweli/kernlog.rs.git", rev = "a44a3f021815e091966741aa3bcdcefdfd51590d" }
+simplelog = { version = "0.8", default-features = false }
+log = "0.4"
 
 [dev-dependencies]
 tempfile = "3"
--- a/TODO
+++ b/TODO
@@ -1,2 +1,1 @@
 - logging in generators needs to go to kmsg, println!() is not enough.
-- read configuration also from /usr/lib?
--- a/TODO
+++ b/TODO
@@ -1,1 +1,0 @@
-- logging in generators needs to go to kmsg, println!() is not enough.
--- a/src/config.rs
+++ b/src/config.rs
@@ -3,6 +3,7 @@
 use anyhow::{anyhow, Context, Result};
 use ini::Ini;
 use liboverdrop::FragmentScanner;
+use log::{info, warn};
 use std::cmp;
 use std::collections::{BTreeMap, HashMap};
 use std::fmt;
@@ -45,7 +46,7 @@ impl Device {
     fn is_enabled(&self, memtotal_mb: u64) -> bool {
         match self.host_memory_limit_mb {
             Some(limit_mb) if limit_mb < memtotal_mb => {
-                println!(
+                info!(
                     "{}: system has too much memory ({:.1}MB), limit is {}MB, ignoring.",
                     self.name,
                     memtotal_mb,
@@ -106,7 +107,7 @@ fn read_devices(root: &Path, memtotal_mb: u64) -> Result<HashMap<String, Device>
     let fragments = locate_fragments(root);
 
     if fragments.is_empty() {
-        println!("No configuration file found.");
+        info!("No configuration file found.");
     }
 
     let mut devices: HashMap<String, Device> = HashMap::new();
@@ -117,9 +118,10 @@ fn read_devices(root: &Path, memtotal_mb: u64) -> Result<HashMap<String, Device>
         for (sname, props) in ini.iter() {
             let sname = match sname {
                 None => {
-                    eprintln!(
-                        "{:?}: ignoring settings outside of section: {:?}",
-                        path, props
+                    warn!(
+                        "{}: ignoring settings outside of section: {:?}",
+                        path.display(),
+                        props
                     );
                     continue;
                 }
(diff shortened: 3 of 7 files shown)


keszybz:
Looks very nice!

Would it be possible to tweak the output to strip the time prefix and [INFO]: right now this is pointlessly duplicated in the journal (and in the wrong timezone):
```
Jun 17 15:56:02 systemd[1666]: Started /home/zbyszek/src/zram-generator/target/debug/zram-generator tests/03-too-much-memory/run/units/.
Jun 17 15:56:02 zram-generator[1254668]: 13:56:02 [INFO] Found configuration for zram0: host-memory-limit=7825MB zram-fraction=0.5 max-zram-size=40>
Jun 17 15:56:02 zram-generator[1254668]: 13:56:02 [INFO] zram0: system has too much memory (7825.4MB), limit is 7825MB, ignoring.
Jun 17 15:56:02 zram-generator[1254668]: 13:56:02 [INFO] No devices configured, exiting.
```

nabijaczleweli:
Hmm it looks like the user you're running the generator as can't write to `/dev/kmsg`, in which case it'll write to stdout with `simplelog` (note how the output you get in the journal is the same as I get as my normal user, in the second part of my screenshot)? The journal output should look more like the top half of the screenshot.

keszybz:
> Hmm it looks like the user you're running the generator as can't write to `/dev/kmsg`

Yep, it was running as user. OK, so maybe it's all OK. The output as unprivileged user is not that imporant, since it'd only be used for testing.

nabijaczleweli:
Rebased. Updated. Log initialisation says what (if anything) went wrong. Failing log init or enabling test mode via `ZRAM_GENERATOR_ROOT` brings loglevel from Info to Trace. See screenshots below.

Manual run
stop & daemon-reload


Preempting @ignatenkobrain: here's the tree diff of with/without simplelog; it ain't bad at all:
```
--- without-simplelog	2020-06-19 18:26:19.702729037 +0200
+++ with-simplelog	2020-06-19 18:26:09.598473393 +0200
@@ -12,7 +12,19 @@
 ├── liboverdrop v0.0.2
 │   └── log v0.4.8 (*)
 ├── log v0.4.8 (*)
-└── rust-ini v0.13.0
+├── rust-ini v0.13.0
+└── simplelog v0.8.0
+    ├── chrono v0.4.11
+    │   ├── num-integer v0.1.43
+    │   │   └── num-traits v0.2.12
+    │   │       [build-dependencies]
+    │   │       └── autocfg v1.0.0
+    │   │   [build-dependencies]
+    │   │   └── autocfg v1.0.0 (*)
+    │   ├── num-traits v0.2.12 (*)
+    │   └── time v0.1.43
+    │       └── libc v0.2.71 (*)
+    └── log v0.4.8 (*)
 [dev-dependencies]
 ├── fs_extra v1.1.0
 └── tempfile v3.1.0
```

nabijaczleweli:
Rebased

keszybz:
LGTM.

Cargo.toml
```diff
@@ -14,6 +14,9 @@ anyhow = "1.0.12"
 clap = { version = "2.33", default-features = false }
 liboverdrop = "0.0.2"
 rust-ini = ">=0.13, <0.16"
+kernlog = { git = "https://github.com/nabijaczleweli/kernlog.rs.git", rev = "a44a3f021815e091966741aa3bcdcefdfd51590d" }
```
ignatenkobrain, line 17: I'm not very comfortable with this because it would mean we are not able to make a release in Fedora...

nabijaczleweli: How about a submodule or a subtree?

ignatenkobrain: Not really, we need it on crates.io

nabijaczleweli: The only options, as I see it, are:
  * speccing a github repository (as it is now),
  * convincing the kernlog maintainer to do a bunch of work (merging the PR, testing, publishing, the lot), or
  * vendoring the thing.

The first one is understandably problematic, the second is either not possible or a bit scummy, I can't really see why the third wouldn't work, unless I'm missing something obvious; it'd potentially duplicate kernlog (though, in reality, kernlog doesn't actually build nowadays), but that's hardly RC?

keszybz: Let's wait for @kstep's response. If the project is dead then my preference would be to simply integrate the code locally. It's ~120 lines of code, and we could probably get rid of the simplelog fallback. So maybe that's the thing to do anyway?

ignatenkobrain: Fully agree, or @nabijaczleweli can just create a new crate that will not do a fallback and we can use that.

nabijaczleweli: >and we could probably get rid of the simplelog fallback

Oh I think I haven't quite explained clearly enough.

As I've noted at the top (reproduced here in full):
1. The maxi approach taken here, using another logger (here: `simplelog`, which is rather large dependency-wise), if kernlog doesn't initialise.
2. The midi approach, using, for example, a "develop" feature, which will enable `simplelog` only for builds you specify (I think one could be devilish and enable it only on the debug build profile, but).
3. The mini approach: simply not. If the logger initialises, great; if not: whatever. If no `log` logger is installed, the output is ignored.
4. The micro approach: `kernlog::init().unwrap()` or equivalent.

Using `simplelog` is just one potential approach that I continued with since nobody objected. Dropping `simplelog` is as trivial as removing the dependency line from `Cargo.toml` and one init line in `main.rs`. To reiterate: it is *not* a property of `kernlog` *at all*.
IOW what you're proposing is simply switching to the mini approach.

I can integrate kernlog into here if you want, but I still think it'd be a better idea to vendor it, so I'll hold off until we're all on the same page.

nabijaczleweli:
New output as non-root, to stdout:
2020-06-26-17:58:52-nabtop-1156x79

nabijaczleweli:
Superseded by #39

PR #33, opened 2020-06-17
Support full set of config paths with dropins

keszybz:


src/config.rs
```diff
@@ -59,37 +86,98 @@ impl fmt::Display for Device {
...
 
-    Ok(Ini::load_from_file(&path)
-        .with_context(|| format!("Failed to read configuration from {}", path.display(),))?
-        .into_iter())
+    let mut devices: HashMap<String, Device> = HashMap::new();
+
+    for (_, path) in fragments {
+        // eprintln!("{}", path.display());
+
+        let ini = Ini::load_from_file(&path).unwrap();
```
nabijaczleweli, line 117: Stray `unwrap()` (also comment)

README.md
```diff
@@ -1,33 +1,45 @@
...
 
-2. Once we figure out all the details, it should be useful as an
-   example of a systemd generator in rust.
+… and the first file found in that list wins.
+
+In addition, "drop-ins" will be loaded from `.conf` files in
+`/etc/systemd/zram-generator.conf.d/`,
+`/usr/lib/systemd/zram-generator.conf.d/`, etc.
+See systemd.unit(5) for a description of this logic.
+
```
lucab, line 31: The journald docs had an explicit line here that stated:
```
The main configuration file is read before any of the configuration directories, and has the lowest precedence;
entries in a file in any configuration directory override entries in the single configuration file.
```

It would be nice to quote at least the first part here too.

src/config.rs
```diff
@@ -59,37 +86,98 @@ impl fmt::Display for Device {
...
-        read_devices(root)?
-            .map(|(sn, s)| parse_device(sn, s, memtotal_mb))
-            .map(Result::transpose)
-            .flatten(),
-    )
+    let memtotal_mb = get_total_memory_kb(&root)? / 1024;
+
+    let devices: Vec<Device> = read_devices(root, memtotal_mb)?
+        .into_iter()
+        .filter(|(_, dev)| dev.disksize > 0)
```
nabijaczleweli, line 98: This is normally disallowed by zram.ko:
```
nabijaczleweli@tarta:~$ sudo tee /sys/block/zram0/disksize
0
0
tee: /sys/block/zram0/disksize: Invalid argument
```

So mentioning that this is an explicit config point "you can disable a device by setting its size to zero" or sth in the README would be nice.

Cargo.toml
```diff
@@ -10,7 +10,7 @@ edition = "2018"
 [dependencies]
 anyhow = "1"
 clap = { version = "2.33", default-features = false }
-rust-ini = "0.13"
+rust-ini = "0.15"
```
ignatenkobrain, line 14: ehh... this will bring quite a lot of dependencies

src/config.rs
```diff
@@ -59,37 +86,98 @@ impl fmt::Display for Device {
...
+        String::from(root.join("usr/lib").to_str().unwrap()),
+        String::from(root.join("usr/local/lib").to_str().unwrap()),
+        String::from(root.join("etc").to_str().unwrap()),
+        String::from(root.join("run").to_str().unwrap()), // We look at /run to allow temporary overriding
+                                                          // of configuration. There is no expectation of
+                                                          // programatic creation of config there.
+    ];
+
+    let cfg = FragmentScanner::new(
+        base_dirs.clone(),
```
nabijaczleweli, line 164: This does four PathBuf allocations, then four String allocations twice, then the vector twice.

The liboverdrop API is quite limiting here, but `base_dirs` can be a `[PathBuf; 4]` (`[root.join("usr/lib"), ...]`), then this argument can be `base_dirs.iter().cloned().map(|pb| pb.into_os_string().into_string().unwrap()).collect()`, which brings it down to four PathBufs, twice, then the vector once.

A comment or `expect()` arg so as to why the unwrap is fine (root is just / in normal operation and otherwise controlled in testing) would be nice as well.

src/config.rs
```diff
@@ -41,6 +41,33 @@ impl Device {
         }
         Ok(())
     }
+
+    fn is_enabled(&self, memtotal_mb: u64) -> bool {
+        if self.host_memory_limit_mb.is_some() && self.host_memory_limit_mb.unwrap() < memtotal_mb {
```
lucab, line 46: This can be written as:
```rust
let enabled = match limit {
    Some(l) if l < total => { println!(...); false },
    _ => true,
};
return enabled;
```

However I'm unsure about the some-and-not-lt case. Does a zero limit mean enabled here?

Cargo.toml
keszybz: Hmm, my thinking was that this will need to be done at some point, and it's better to update before adding new code that touches the same paths. What would the plan be? To always keep on the old version?

ignatenkobrain: I wanted to bring upstream attention to this problem, but did not have time yet.

src/config.rs
lucab: We are free to adjust the liboverdrop API, it isn't stellar about this right now. Signature suggestions (and PRs) welcome.

src/config.rs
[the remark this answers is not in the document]
keszybz: My thinking was that at some point we'll want to add other conditions here. And then the pattern with `some` will not work. But I'll change it to the suggested for for now.

src/config.rs
keszybz: Hmm, but does it matter? 4 allocations like this are completely negligible, and we do it just once.

src/config.rs
keszybz: > Does a zero limit mean enabled here?

0 means disabled, None means enabled.

keszybz:
Updated to address the review comments above (except as noted).

Cargo.toml
ignatenkobrain: given that it is essentially drop-in replacement with your PR (I mean changing to 0.13 still compiles and tests pass), I think it is fine to keep whatever you like. In Fedora we can just revert this version bump.

ignatenkobrain:
```
└── rust-ini v0.15.3
    ├── cfg-if v0.1.10
    └── ordered-multimap v0.2.4
        ├── dlv-list v0.2.2
        │   └── rand v0.7.3
        │       ├── getrandom v0.1.14
        │       │   ├── cfg-if v0.1.10
        │       │   └── libc v0.2.71
        │       ├── libc v0.2.71
        │       ├── rand_chacha v0.2.2
        │       │   ├── ppv-lite86 v0.2.8
        │       │   └── rand_core v0.5.1
        │       │       └── getrandom v0.1.14 (*)
        │       └── rand_core v0.5.1 (*)
        └── hashbrown v0.7.2
            └── ahash v0.3.8
            [build-dependencies]
            └── autocfg v1.0.0
```

It is basically this big list of dependencies. That is twice bigger than all other dependencies and their recursive dependencies too.

src/config.rs
nabijaczleweli: It might not matter for the purposes of performance, but it's very very wordy and it was hard to decipher what the code was *trying* to do. Consider this implementation instead, which does away with most of the fluff (fewer allocations are a bonus) and is easier to figure out:
```rust
fn locate_fragments(root: &Path) -> BTreeMap<String, PathBuf> {
    let mut base_dirs = [
        root.join("usr/lib"),
        root.join("usr/local/lib"),
        root.join("etc"),
        root.join("run"), // We look at /run to allow temporary overriding
                          // of configuration. There is no expectation of
                          // programatic creation of config there.
    ];

    let cfg = FragmentScanner::new(
        base_dirs.iter().cloned().map(|pb| pb.into_os_string().into_string().unwrap()).collect(),
        "systemd/zram-generator.conf.d",
        true,
        vec![String::from("conf")],
    );

    let mut fragments = cfg.scan();

    for path in base_dirs.iter_mut().rev() {
        path.push("systemd/zram-generator.conf");
        if path.exists() {
            fragments.insert(String::new(), path); // The empty string shall sort earliest
            break;
        }
    }

    fragments
}
```

src/config.rs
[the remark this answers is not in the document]
nabijaczleweli: In that case evolving that into something like, e.g.
```rust
match self.host_memory_limit_mb {
    Some(limit_mb) if limit_mb < memtotal_mb => {
        // code
        return false;
    }
    _ => {},
}

// more clauses

true
```
would still be a win readability-wise IMO

src/config.rs
ignatenkobrain: Can you submit this as a separate PR please?

nabijaczleweli: Always: #34

PR #34, opened 2020-06-17
Simplify config::locate_fragments()

nabijaczleweli:
As seen on https://github.com/systemd/zram-generator/pull/33#discussion_r441784913

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -8,6 +8,7 @@ use std::collections::{BTreeMap, HashMap};
 use std::fmt;
 use std::fs;
 use std::io::{prelude::*, BufReader};
+use std::mem;
 use std::path::{Path, PathBuf};
 
 pub struct Device {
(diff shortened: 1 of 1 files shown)


src/config.rs
```diff
@@ -149,28 +150,32 @@ fn read_devices(root: &Path, memtotal_mb: u64) -> Result<HashMap<String, Device>
...
+                          // programatic creation of config there.
     ];
 
     let cfg = FragmentScanner::new(
-        base_dirs.clone(),
+        base_dirs
+            .iter()
+            .cloned()
+            .map(|pb| pb.into_os_string().into_string().unwrap())
+            .collect(),
```
keszybz, line 167: I'm not convinced that this a net gain. The original code was relatively simple, and was using immutable data structures. I don't think saving a few memory allocations is important enough to add such complicated code and use mutable variables.

nabijaczleweli: Fair enough. I've since opened https://github.com/overdrop/liboverdrop-rs/pull/11, which will make most of the additional hard-type complexity here redundant. I'll revisit this if/when that gets merged, but I'll close it for now.

PR #35, opened 2020-06-17
Bump min version of anyhow & bump version to 0.2.0-beta.1

ignatenkobrain:
Even though tests are not working in some build environments (like Fedora), it would be still good to ship a new version ASAP so that people can test it.

We will fix the issue in release.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "zram-generator"
-version = "0.1.2"
+version = "0.2.0-beta.1"
 authors = ["Zbigniew Jędrzejewski-Szmek <zbyszek@in.waw.pl>"]
 license = "MIT"
 description = "Systemd unit generator for zram swap devices"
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -1,14 +1,14 @@
...
+version = "0.2.0-beta.1"
 authors = ["Zbigniew Jędrzejewski-Szmek <zbyszek@in.waw.pl>"]
 license = "MIT"
 description = "Systemd unit generator for zram swap devices"
 homepage = "https://github.com/systemd/zram-generator"
 edition = "2018"
 
 [dependencies]
-anyhow = "1"
+anyhow = "1.0.12"
```
nabijaczleweli, line 11: Doesn't this hardlock anyhow to 1.0.12? Something like `>=1.0.12` should be more like what you intended

ignatenkobrain: No, `1.0.12` means `^1.0.12` that essentially means `>=1.0.12, <2.0.0`

ignatenkobrain: Just using `>=1.0.12` is wrong because even 2.x, 3.x and such will match this requirement.

PR #38, opened 2020-06-19
Manpages

### LINKED ISSUE #22: Manpage(s)?
I [edited](https://github.com/nabijaczleweli/systemd-zram-generator.deb/blob/882f88da4ae0984d914e6d7de9d464245179ecc4/debian/systemd-zram-generator.8) a manpage from a different generator for my package because I felt bad packaging something with *no* manpage at all, but, well, something that's three lines long and doesn't document the config file (you could argue the [default config file](https://github.com/nabijaczleweli/systemd-zram-generator.deb/blob/53029c4e82fa48871ab437eefea993fd188c62de/debian/zram-generator.conf) is documentation enough I guess) is a step-down from the rest of systemd-adjacent documentation.

Quoth the Policy, [section 12.1 "Manual pages"](https://www.debian.org/doc/debian-policy/ch-docs.html):
> You should install manual pages in nroff source form, in appropriate places under /usr/share/man. You should only use sections 1 to 9 (see the FHS for more details). You must not install a pre-formatted “cat page”.
>
> Each program, utility, and function should have an associated manual page included in the same package. It is suggested that all configuration files also have a manual page included as well. Manual pages for protocols and other auxiliary things are optional.
>
> If no manual page is available, this is considered as a bug and should be reported to the Debian Bug Tracking System (the maintainer of the package is allowed to write this bug report themselves, if they so desire). Do not close the bug report until a proper man page is available.

I'm not sure (and, frankly, don't much care) about fedora policy, but fedora isn't crux, and apparently this is gonna be a Point of a subsequent release, so I'm assuming having some documentation would be nice.

I'd be open to writing said manpages myself, but I don't know what your preferred generator is/would be for this (my personal preference is `ronn`, which outputs [man(1)](https://rawcdn.githack.com/LoungeCPP/pir-8-emu/man/pir-8-emu.1) and [HTML](https://rawcdn.githack.com/LoungeCPP/pir-8-emu/man/pir-8-emu.1.html) pages from a [markdown dialect](https://github.com/LoungeCPP/pir-8-emu/blob/master/man/pir-8-emu.md)).

nabijaczleweli:
Plopping this here before I crash for tonight for broad strokes preprereview.

Build with `ronn man/*.md`.

An unnamed notes tab in my editor says:
```
style: `TARGET_DIR` vs (*TARGET_DIR* + other styling for other things)
```

Closes #22

### DIFF
--- a/man/zram-generator.conf.md
+++ b/man/zram-generator.conf.md
@@ -1,0 +1,112 @@
+zram-generator.conf(5) -- Systemd unit generator for zram swap devices (configuration)
+======================================================================================
+
+## SYNOPSIS
+
+`/usr/lib/systemd/zram-generator.conf`<br />
+`/usr/local/lib/systemd/zram-generator.conf`<br />
+`/etc/systemd/zram-generator.conf`<br />
+`/run/systemd/zram-generator.conf`
+
+`/usr/lib/systemd/zram-generator.conf.d/*.conf`<br />
+`/usr/local/lib/systemd/zram-generator.conf.d/*.conf`<br />
+`/etc/systemd/zram-generator.conf.d/*.conf`<br />
+`/run/systemd/zram-generator.conf.d/*.conf`
+
+## DESCRIPTION
+
+These files configure devices created by zram-generator(8). See systemd.syntax(5) for a general description of the syntax.
+
+## CONFIGURATION DIRECTORIES AND PRECEDENCE
+
+The default configuration doesn't specify any devices. Consult */usr/share/zram-generator/zram-generator.conf.example* for an example configuration file.
+
+When packages need to customize the configuration, they can install configuration snippets in */usr/lib/systemd/zram-generator.conf.d/*.
+Files in */etc/* are reserved for the local administrator, who may use this logic to override the configuration files installed by vendor packages.
+The main configuration file is read before any of the configuration directories, and has the lowest precedence;
+entries in a file in any configuration directory override entries in the single configuration file.
+Files in the *\*.conf.d/* configuration subdirectories are sorted by their filename in lexicographic order, regardless of which of the subdirectories they reside in.
+When multiple files specify the same option, for options which accept just a single value, the entry in the file with the lexicographically latest name takes precedence.
+It is recommended to prefix all filenames in those subdirectories with a two-digit number and a dash, to simplify the ordering of the files.
+
+To disable a configuration file supplied by the vendor, the recommended way is to place a symlink to */dev/null* in the configuration directory in */etc/*,
+with the same filename as the vendor configuration file.
+
+## OPTIONS
+
+Each device is configured independently in its `[zramN]` section, where N is a nonnegative integer. Other sections are ignored.
+
+Devices with the final size of *0* will be discarded.
+
+* `host-memory-limit`=
+
+  Sets the upper limit on the total usable RAM (as defined by *MemTotal* in `/proc/meminfo`, confer proc(5)) above which the device will *not* be created.
+
+  This takes a nonnegative number, representing that limit in megabytes, or the literal string *none*, which can be used to override a limit set earlier.
+
+  Defaults to *none*.
+
+  For compatibility with earlier versions, `memory-limit` is allowed as an alias for this option.
+  Its use is discouraged, and administrators should migrate to `host-memory-limit`.
+
+* `zram-fraction`=
+
+  Defines the scaling factor of the zram device's size with relation to the total usable RAM.
+
+  This takes a nonnegative floating-point number representing that factor.
+
+  Defaults to *0.5*.
+
+* `max-zram-size`=
+
+  Sets the limit on the zram device's size obtained by `zram-fraction`.
+
+  This takes a nonnegative number, representing that limit in megabytes, or the literal string *none*, which can be used to override a limit set earlier.
+
+  Defaults to *4096*.
+
+* `compression-algorithm`=
+
+  Specifies the algorithm used to compress the zram device.
+
+  This takes a literal string, representing the algorithm to use.<br />
+  Consult */sys/block/zram0/comp_algorithm* for a list of currently loaded compression algorithms, but note that additional ones may be loaded on demand.
+
+  If unset, none will be configured and the kernel's default will be used.
+
+## ENVIRONMENT VARIABLES
+
+Setting `ZRAM_GENERATOR_ROOT` during parsing will cause */proc/meminfo* to be read from *$ZRAM_GENERATOR_ROOT/proc/meminfo* instead,
+and *{/usr/lib,/usr/local/lib,/etc,/run}/systemd/zram-generator.conf* to be read from *$ZRAM_GENERATOR_ROOT/{/usr/lib,/usr/local/lib,/etc,/run}/systemd/zram-generator.conf*.
+
+## EXAMPLES
+
+The default configuration will yield the following:
+
+     zram device size [MB]
+         ^
+         │
+      4G>│               ooooooooooooo
+         │             o
+         │           o
+         │         o
+      2G>│       o
+         │     o
+         │   o
+    512M>│ o
+         0───────────────────────> total usable RAM [MB]
+           ^     ^       ^
+           1G    4G      8G
+
+## REPORTING BUGS
+
+&lt;<https://github.com/systemd/zram-generator/issues>&gt;
+
+## SEE ALSO
+
+zram-generator(8), systemd.syntax(5), proc(5)
+
+&lt;<https://github.com/systemd/zram-generator>&gt;
+
+Linux documentation of zram: &lt;<https://kernel.org/doc/html/latest/admin-guide/blockdev/zram.html>&gt;<br />
+     and the zram sysfs ABI: &lt;<https://kernel.org/doc/Documentation/ABI/testing/sysfs-block-zram>&gt;
--- a/man/zram-generator.md
+++ b/man/zram-generator.md
@@ -1,0 +1,50 @@
+zram-generator(8) -- Systemd unit generator for zram swap devices
+=================================================================
+
+## SYNOPSIS
+
+`/usr/lib/systemd/system-generators/zram-generator` `TARGET_DIR` [*2RGET_DIR* *3RGET_DIR*]<br />
+`/usr/lib/systemd/system-generators/zram-generator` --setup-device `DEVICE`<br />
+`/usr/lib/systemd/system-generators/zram-generator` --reset-device `DEVICE`
+
+## DESCRIPTION
+
+`zram-generator` is a generator that creates zram devices and systemd units to configure, format, and enable swap on them.
+
+
+The generator will be invoked by systemd early at boot. The generator will then:
+
+  1. read configuration files from *{/etc,/lib}/systemd/zram-generator.conf[.d]* (see zram-generator.conf(5) for details);
+  2. generate systemd.swap(5) units into `TARGET_DIR` and make `swap.target` depend on them;
+  3. ensure the `zram` module is loaded and create the requested devices.
+
+The generator does nothing if run inside a container (as determined by *systemd-detect-virt(8) --container*).
+
+Setting the `ZRAM_GENERATOR_ROOT` environment variable makes the generator run in test mode, in which case containerisation is ignored and step `3` is skipped.<br />
+For the ramifications of `ZRAM_GENERATOR_ROOT` on config handling, see zram-generator.conf(5).
+
+
+The generated *dev-zramN.swap* units depend on `zram-create@zramN.service`, which will:
+
+  1. read configuration files from *{/etc,/lib}/systemd/zram-generator.conf[.d]* (see zram-generator.conf(5) for details);
+  2. set the desired compression algorithm, if any;
+     if the current kernel doesn't understand the specified algorithm, a warning is issued, but execution continues;
+  3. set the desired blockdev size and format it as swap with *systemd-makefs(8)*.
+
+When the unit is stopped, the zram device is reset, freeing memory and allowing the device to be reused.
+
+
+`zram-generator` implements systemd.generator(7).
+
+## REPORTING BUGS
+
+&lt;<https://github.com/systemd/zram-generator/issues>&gt;
+
+## SEE ALSO
+
+zram-generator.conf(5), systemd.generator(7), systemd.swap(5)
+
+&lt;<https://github.com/systemd/zram-generator>&gt;
+
+Linux documentation of zram: &lt;<https://kernel.org/doc/html/latest/admin-guide/blockdev/zram.html>&gt;<br />
+     and the zram sysfs ABI: &lt;<https://kernel.org/doc/Documentation/ABI/testing/sysfs-block-zram>&gt;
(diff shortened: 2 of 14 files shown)


man/zram-generator.md
```diff
@@ -0,0 +1,57 @@
+zram-generator(8) -- Systemd unit generator for zram swap devices
+=================================================================
+
+## SYNOPSIS
+
+`/lib/systemd/system-generators/zram-generator` `TARGET_DIR` [*2RGET_DIR* *3RGET_DIR*]<br />
+`/lib/systemd/system-generators/zram-generator` --setup-device `DEVICE`<br />
+`/lib/systemd/system-generators/zram-generator` --reset-device `DEVICE`
```
keszybz, line 8: `/usr/lib` please. Both versions are equivalent on modern systems, but `/usr/lib` is the canonical location.

man/zram-generator.md
```diff
@@ -0,0 +1,57 @@
...
+
+## DESCRIPTION
+
+`zram-generator` is a generator that creates zram devices and systemd units to configure, format, and enable swap on them.
+
+The generator will be invoked by systemd early at boot, the generator will then:
```
keszybz, line 14: "the generator will then" is a separate sentence. Change "," to ".".

man/zram-generator.md
```diff
@@ -0,0 +1,57 @@
...
+
+  1. read configuration files from *{/etc,/lib}/systemd/zram-generator.conf[.d]* (see zram-generator.conf(5) for details);
+  2. if systemd-detect-virt(1) exists and returns non-*0*, exit, since continuing would mean changing the host system;
+  3. generate systemd.swap(5) units into `TARGET_DIR` for all configured devices and symlink them into *swap.target.wants/*;
+  4. if any devices are enabled, and the zram module isn't loaded (*/sys/class/zram-control* doesn't exist), invoke *modprobe(8) zram*;
```
keszybz, line 19: I don't think we want to describe this with this level of detail. The implementation is likely to change, and then we'd have to update the man page each time. Maybe just say "generate appropriate units"?

man/zram-generator.md
```diff
@@ -0,0 +1,57 @@
...
+  5. if the requested devices aren't present, read */sys/class/zram-control/hot_add* until the highest device's number is read,
```
keszybz, line 20: Same for 4–5: maybe just "The generator will try to ensure that the `zram` module is loaded.".

man/zram-generator.md
```diff
@@ -0,0 +1,57 @@
...
+`/lib/systemd/system-generators/zram-generator` --reset-device `DEVICE`
+
+## DESCRIPTION
+
+`zram-generator` is a generator that creates zram devices and systemd units to configure, format, and enable swap on them.
+
+The generator will be invoked by systemd early at boot, the generator will then:
+
+  1. read configuration files from *{/etc,/lib}/systemd/zram-generator.conf[.d]* (see zram-generator.conf(5) for details);
+  2. if systemd-detect-virt(1) exists and returns non-*0*, exit, since continuing would mean changing the host system;
```
keszybz, line 17: Implementation detail... I'd prefer to say, outside of this list, that "The generator does nothing in containers, except when running in test mode, in particular when $ZRAM_GENERATOR_ROOT is set." (The distinction about containers is important, since it *does* work on VMs.)

man/zram-generator.md
```diff
@@ -0,0 +1,57 @@
...
+The generated *dev-zramN.swap* units depend on `zram-create@zramN.service`, which on start invokes the generator with *--setup-device zramN*, which will then:
+
+  1. read configuration files from *{/etc,/lib}/systemd/zram-generator.conf[.d]* (see zram-generator.conf(5) for details);
+  2. set the desired compression algorithm in */sys/block/zramN/comp_algorithm*, if any;
+     if the write fails with `EINVAL` the current kernel doesn't understand the specified algorithm; a warning is issued, but execution continues as normal,
+     if the write fails for any other reason, setup is halted;
+  3. set the desired blockdev size in */sys/block/zramN/disksize*, if the write fails, setup is halted;
+  4. run *systemd-makefs(8) swap /dev/zramN*
+
+And on stop with *--reset-device zramN*, which will write a positive integer to */sys/block/zramN/reset*, allowing the device to be reused.
```
keszybz, line 35: I'd remove this paragraph and the previous one. Just describe the effect of the unit: the zram device is set up according to configuration, formatted as a swap device, and enabled. When the unit is stopped, the zram device is reset to free memory.

man/zram-generator.md
```diff
@@ -0,0 +1,57 @@
...
+
+
+`zram-generator` implements systemd.generator(7).
+
+## AUTHOR
```
keszybz, line 40: "S"

man/zram-generator.md
```diff
@@ -0,0 +1,57 @@
...
+
+&lt;<https://github.com/systemd/zram-generator/issues>&gt;
+
+## SEE ALSO
+
+zram-generator.conf(5), systemd.generator(7)
+
+&lt;<https://github.com/systemd/zram-generator>&gt;
+
+Linux documentation of zram: &lt;<https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/admin-guide/blockdev/zram.rst?h=v5.7>&gt;<br />
```
keszybz, line 56: https://www.kernel.org/doc/html/latest/admin-guide/blockdev/zram.html

man/zram-generator.md
```diff
@@ -0,0 +1,57 @@
...
+     and the zram sysfs ABI: &lt;<https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/ABI/testing/sysfs-block-zram?h=v5.7>&gt;
```
keszybz, line 57: Is this also rendered somewhere? We should link to html and to the latest version.

ignatenkobrain: I think here: https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-block-zram

man/zram-generator.md
nabijaczleweli: Oh, cool. I've made it a versioned link though, since these things can and do move (it's in Documentation/blockdev/zram.rst on the 4.19 buster kernel, but moved to admin-guide/ sometime since), and usually nobody notices. Indeed, I recently fixed a whole host of these in the util-linux manpages which nobody noticed since 2016.

man/zram-generator.md
nabijaczleweli: Updated

man/zram-generator.md
nabijaczleweli: Updated

man/zram-generator.md
nabijaczleweli: Oh, good catch! Updated

man/zram-generator.md
nabijaczleweli: I've simplified this sexion, but left the broad-strokes blow-by-blow, since it highlights (a) the unrecognised compression behaviour and (b) the potential race between generation and setup (e.g. if swap.target is started manually later).

man/zram-generator.md
nabijaczleweli: I've simplified this a bit, but I've left the `TARGET_DIR` intact, since it's highlit in the SYNOPSIS (and this reinforces that we don't use the 2rget or 3rget), and said "make `swap.target` depend on them", which i think is the happiest marriage (anything more is implementation detail, anything less is unclear re: how (if) these are pulled in and when)

nabijaczleweli:
Added `zram-generator.conf(5)`; the intro is based on `journald.conf`, might need further cutting down.

src/config.rs
```diff
@@ -199,7 +198,14 @@ fn parse_line(dev: &mut Device, key: &str, value: &str) -> Result<()> {
         "zram-fraction" => {
```
nabijaczleweli, line 198: @keszybz: you implemented the thing above:
```rust
    for dir in base_dirs.iter().rev() {
        let path = PathBuf::from(dir).join("systemd/zram-generator.conf");
        if path.exists() {
            fragments.insert("".to_string(), path); // The empty string shall sort earliest
            break;
        }
    }
```

This feels broken now that I look at it: `fragments.insert()` will replace all config files with the latest one it finds (in `/usr/lib`). Is this what you intended?

ignatenkobrain: IIUC yes. it has rev() so it should go from /usr down to /run so the first one wins. aka /etc will override /usr/lib completely.

nabijaczleweli: Given
```rust
    let base_dirs = vec![
        String::from(root.join("usr/lib").to_str().unwrap()),
        String::from(root.join("usr/local/lib").to_str().unwrap()),
        String::from(root.join("etc").to_str().unwrap()),
        String::from(root.join("run").to_str().unwrap()), // We look at /run to allow temporary overriding
                                                          // of configuration. There is no expectation of
                                                          // programatic creation of config there.
    ];
```

`base_dirs.rev()` will yield `["/usr/lib", "/usr/local/lib", "/etc", "/run"]`

nabijaczleweli: Given
```rust
    let base_dirs = vec![
        String::from(root.join("usr/lib").to_str().unwrap()),
        String::from(root.join("usr/local/lib").to_str().unwrap()),
        String::from(root.join("etc").to_str().unwrap()),
        String::from(root.join("run").to_str().unwrap()), // We look at /run to allow temporary overriding
                                                          // of configuration. There is no expectation of
                                                          // programatic creation of config there.
    ];
```

`base_dirs.rev()` will yield `["/run", "/etc", "/usr/local/lib", "/usr/lib"]` => `/usr/lib/...`, if found, will override `/run/...`, if found

nabijaczleweli: Ugh sorry disregard this, I missed the `break;`.

man/zram-generator.md
```diff
@@ -0,0 +1,56 @@
...
+
+&lt;<https://github.com/systemd/zram-generator/issues>&gt;
+
+## SEE ALSO
+
+zram-generator.conf(5), systemd.generator(7), systemd.swap(5)
+
+&lt;<https://github.com/systemd/zram-generator>&gt;
+
+Linux documentation of zram: &lt;<https://kernel.org/doc/html/v5.6/admin-guide/blockdev/zram.html>&gt;<br />
```
ignatenkobrain, line 55: Probably https://www.kernel.org/doc/html/latest/admin-guide/blockdev/zram.html instead?

zram-generator.conf.example
```diff
@@ -6,14 +6,11 @@
 #
 # "host-memory-limit = none" may be used to disable this limit. This
 # is also the default.
-#
-# To disable a device (when some configuration for it already exits),
-# set "host-memory-limit = 0".
```
ignatenkobrain, line 11: hmm, why do we remove this?

zram-generator.conf.example
```diff
@@ -6,14 +6,11 @@
...
 host-memory-limit = 9048
 
 # The fraction of memory to use as ZRAM. For example, if the machine
 # has 1 GiB, and zram-fraction=0.25, then the zram device will have
-# 256 MiB. Values in the range 0.10–0.40 are recommended.
+# 256 MiB. Values in the range 0.10–0.50 are recommended.
 #
 # The default is 0.5.
 zram-fraction = 0.10
```
Maryse47, line 16: Could you match defaults here and elsewhere in example config? I think this will be more intuitive than some random values and also more ready to use example config as current one without having to tweak it.

ignatenkobrain: The example config is much more simpler than this one: https://src.fedoraproject.org/rpms/rust-zram-generator/blob/master/f/zram-generator.conf

man/zram-generator.md
nabijaczleweli: As noted above (https://github.com/systemd/zram-generator/pull/38#discussion_r442776280): these things move and nobody notices and then the links are dead. This is best avoided.

zram-generator.conf.example
nabijaczleweli: Misleading: anything that will lead to disksize=0 will disable a device. This is just about the noisiest method for doing so, and max size/fraxion are better for this.

zram-generator.conf.example
nabijaczleweli: Similarly, the config I bundle with my Debian package does actually match: https://github.com/nabijaczleweli/systemd-zram-generator.deb/blob/debian/debian/zram-generator.conf. Distributor's choice v0v

Maryse47: My point was that the upstream example is filled with arbitrary values which are more or less far from optimal ones which some users may treat as ready to use recommendations. Debian version looks much better in this regard and I believe it would serve as upstream example better otherwise this may end up with every distro shipping their own example which would be a unnecessary duplication of work and waste.

man/zram-generator.md
keszybz: I'd prefer to keep the "dynamic" link. There's nothing specific about kernel 5.7, and we want to always refer to the latest version of the kernel documentation. If the link becomes stale, it'll be easy to fix. (Though, I hope that the kernel learns to do redirects for link stability so this won't be necessary.)

nabijaczleweli: Done

man/zram-generator.conf.md
```diff
@@ -0,0 +1,118 @@
...
+    512M>│ o
+         0───────────────────────> total usable RAM [MB]
+           ^     ^       ^
+           1G    4G      8G
+
+## AUTHORS
+
+Written by Zbigniew Jędrzejewski-Szmek &lt;<zbyszek@in.waw.pl>&gt;,
+           Igor Raits &lt;<i.gnatenko.brain@gmail.com>&gt;,
+       and наб &lt;<nabijaczleweli@gmail.com>&gt;
```
keszybz, line 105: Hmm, maybe drop the AUTHORS section? Those tend to get out of date pretty quickly... And then there's always the issue where the cutoff point should be. We now have 7 authors by `git log`, and soon there'll be more. So I think it's OK to keep a short list in Cargo.toml, but I'd just stick to that.

man/zram-generator.conf.md
```diff
@@ -0,0 +1,118 @@
...
+  This takes a nonnegative number, representing that limit in megabytes, or the literal string *none*, which can be used to override a limit set earlier.
+
+  Defaults to *4096*.
+
+* `compression-algorithm`=
+
+  Specifies the algorithm used to compress the zram device.
+
+  This takes a literal string, representing the algorithm to use.<br />
+  See */sys/block/zram0/comp_algorithm* for the list of algorithms supported by the currently running kernel.
```
keszybz, line 73: This is a bit problematic. That file shows a list of *currently loaded* compression algorithms, and further algs may be loaded if available when a string is written here. So maybe say 
`*/sys/block/zram0/comp_algorithm* contains a list of currently loaded compression algorithms. Further algorithms may be loaded on demand.`

nabijaczleweli: Updated.

man/zram-generator.conf.md
nabijaczleweli: Fair enough; removed the list from both manpages and the note in Cargo.toml.

nabijaczleweli:
I had also wanted to add manpage autodeployment to the CI, but I don't have neither the right privileges to add the tokens, nor any experience with GH Actions, so I'll leave it to someone with both.

Issue #40, opened 2020-07-06
`make install` fails on Debian


pdewacht:
I was inspired to try zram-generator after it was mentioned in [LWN](https://lwn.net/SubscriberLink/824620/f3c3b8f9ef9fd704/). But I ran into some problems when installing it:

- `make install` installed to `/usr/lib/systemd/system-generator`, but systemd on Debian does not search that directory

- I could override the installation location by setting the `SYSTEMD_DIR` makefile variable, but `swap-create@.service` still hardcoded the wrong location and I had to edit that file by hand.

A possible fix is to change the default `SYSTEMD_DIR` to `/usr/local/lib/systemd/`. That should work on all systems, if I read the systemd docs correctly.

Alternatively the  system directory could be extracted from pkg-config (something like `pkg-config systemd --variable systemdsystemgeneratordir`?)

pdewacht:
Also, `src/main.rs` hardcodes `/usr/lib/systemd/systemd-makefs`, which is also not correct on Debian.

nabijaczleweli:
Which is why I mass-replace `/usr/lib/systemd` with `/lib/systemd` in my Debian package: https://github.com/nabijaczleweli/systemd-zram-generator.deb/blob/a9e5600442908e110e7f8668d9a13ffa5dad2387/debian/rules#L12. Chalk it up as a distro-specific patch v0v

Merged-usr is the default on new Debian installs, and merged-usr systems are not affected for obvious reasons. Cargo doesn't really support setting arbitrary strings at build time (which is to say you *could* do it, but it's more trouble than is worth IMO).

PR #41, opened 2020-07-07
Use pkg-config for systemd directories

pdewacht:
This is a fix for the installation problems on Debain, #40. (Or not just Debian, probably all non-merged-/usr distributions are affected?)

----

Add pkg-config calls to the Makefile.

Use an environment variable to pass the location of systemd-makefs to
the Rust code. This gets evaluated at compile-time. If the env var is
unset (e.g. when cargo is invoked by hand), a default is used.

The unit file also needs to be edited before installation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,13 +2,22 @@ INSTALL = install
 CARGO = cargo
 RONN = ronn
 PREFIX = /usr
-SYSTEMD_DIR ?= $(PREFIX)/lib/systemd
+
+SYSTEMD_UTIL_DIR := $(shell pkg-config --variable=systemdutildir systemd)
+SYSTEMD_SYSTEM_UNIT_DIR := $(shell pkg-config --variable=systemdsystemunitdir systemd)
+SYSTEMD_SYSTEM_GENERATOR_DIR := $(shell pkg-config --variable=systemdsystemgeneratordir systemd)
+
+SYSTEMD_MAKEFS_COMMAND = $(SYSTEMD_UTIL_DIR)/systemd-makefs
+export SYSTEMD_MAKEFS_COMMAND
 
 .DEFAULT: build
 .PHONY: build man check clean install
 
 build:
 	@$(CARGO) build --release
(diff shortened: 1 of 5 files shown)


src/setup.rs
```diff
@@ -9,6 +9,9 @@ use std::os::unix::process::ExitStatusExt;
 use std::path::Path;
 use std::process::Command;
 
+const SYSTEMD_MAKEFS_COMMAND: Option<&str> = std::option_env!("SYSTEMD_MAKEFS_COMMAND");
```
nabijaczleweli, line 12: Something like
```rust
const SYSTEMD_MAKEFS_COMMAND: &str = concat!(env!("SYSTEMD_UTIL_DIR"), "/systemd-makefs");
// or even
const SYSTEMD_MAKEFS_COMMAND: &str = concat!(env!("SYSTEMD_UTIL_DIR", "Define $SYSTEMD_UTIL_DIR to the result of $(pkg-config --variable=systemdutildir systemd) (e.g. /usr/lib/systemd/)"), "/systemd-makefs");
```
would be better, the duplication of logic between here and the makefile could easily go out of sync.

Makefile
```diff
@@ -2,13 +2,22 @@ INSTALL = install
 CARGO = cargo
 RONN = ronn
 PREFIX = /usr
-SYSTEMD_DIR ?= $(PREFIX)/lib/systemd
+
+SYSTEMD_UTIL_DIR := $(shell pkg-config --variable=systemdutildir systemd)
```
nabijaczleweli, line 6: `pkg-config` might want to be configured like `cargo` or `ronn` as well

nabijaczleweli:
The README is also untouched and says, i.a. "`units/swap-create@.service` copied into `/usr/lib/systemd/system/`", which is a patent lie.

nabijaczleweli:
Consider using 6d0e8589f1f26d97bf1b520b5b64c72ab7609d77 from my [`touché`](https://github.com/nabijaczleweli/systemd-zram-generator/tree/touch%C3%A9) branch instead of manually defining that env var

pdewacht:
Ah yes, I'll use that. I was going to suggest making that myself (I didn't know enough about this GitHub stuff to do it myself)

keszybz:
LGTM.

PR #43, opened 2020-07-09
Add option in /proc/cmdline to enable/disable the generator

### LINKED ISSUE #42: RFE: parse boot parameters for an Anaconda inbibit parameter
Currently Anaconda has its own built-in swap-on-zram implementation, which can be inhibited by 'inst.zram'. Since the plan is for that implementation to be obsoleted in favor of zram-generator, it might be nice to keep this option functional.

Igor's idea is for the generator to parse for 'inst.zram' or 'inst.nozram' and simply inhibit. And have anaconda requires 'zram-generator-defaults' so that it can by default depend on swap-on-zram being available *unless* inhibited by this option. That seems more elegant than my idea that Anaconda go to the trouble of stopping the service created by the generator.
https://www.redhat.com/archives/anaconda-devel-list/2020-July/msg00016.html

keszybz:
Closes #42.


src/config.rs
```diff
@@ -252,6 +252,27 @@ fn get_total_memory_kb(root: &Path) -> Result<u64> {
     _get_total_memory_kb(&path)
 }
 
+fn _kernel_has_option(path: &Path, word: &str) -> Result<bool> {
+    let text = fs::read_to_string(path)?;
+
+    // The last argument wins, so check all words in turn.
+    let mut val = false;
```
nabijaczleweli, line 259: This is more obvious and doesn't allocate four times per each item
```rust
    // The last argument wins, so check all words in turn.
    Ok(text.split_whitespace().fold(false, |acc, w| {
        if !w.starts_with(word) {
            acc
        } else {
            match &w[word.len()..] {
                "" | "=1" | "=on" => true,
                "=0" | "=off" => false,
                _ => acc,
            }
        }
    }))
```

keszybz: I amended the patch with that.

Anyway, we need to wait for a reply from the anaconda folks whether they want this at all.

cmurf:
Anaconda folks have decided to deprecate inst.zram and leave it up to zram-generator folks if there's a use case for command line inhibition.
https://www.redhat.com/archives/anaconda-devel-list/2020-July/msg00025.html

ignatenkobrain:
In that case, I think it is not worth implementing some option to disable it.

cmurf:
> In that case, I think it is not worth implementing some option to disable it.

Yeah, I think its "defaultness" (the state of whether it is or is not enabled by default) is so fundamental to the image in question that it must be correct.

keszybz:
I think it'd be worthwhile to have a kernel-commandline option. I'm not sure about the name though. `systemd.zram=0|1` ?

src/main.rs
```diff
@@ -67,6 +68,15 @@ fn main() -> Result<()> {
 
     match get_opts() {
         Opts::GenerateUnits(target) => {
+            match config::kernel_has_option(root, "inst.zram") {
+                Ok(true) => {
+                    info!("Disabled by inst.zram option in /proc/cmdline, exiting.");
+                    return Ok(());
+                }
+                Ok(false) => {}
+                Err(_) => warn!("Failed to parse /proc/cmdline, ignoring."),
```
nabijaczleweli, line 77: How about something like this?
```rust
Err(e) => warn!("Failed to read /proc/cmdline ({}), ignoring.", e),
```

nabijaczleweli:
`systemd.zram=` sounds good

keszybz:
Updated with `systemd.zram[=0|1]`.

man/zram-generator.md
```diff
@@ -20,6 +20,8 @@ The generator will be invoked by systemd early at boot. The generator will then:
 
 The generator does nothing if run inside a container (as determined by *systemd-detect-virt(8) --container*).
 
+The generator also understands the kernel command-line option `systemd.zram`. See zram-generator.conf(8) for details.
```
nabijaczleweli, line 23: (5)

src/main.rs
```diff
@@ -65,14 +65,22 @@ fn main() -> Result<()> {
 
     let _ = kernlog::init_with_level(log_level);
 
+    let kernel_override = || match config::kernel_zram_option(root) {
+        Some(false) => std::process::exit(0),
```
nabijaczleweli, line 69: An `{info!("Disabled by kernel cmdline option"); return Ok(())}` would be better usability-wise.

man/zram-generator.md
keszybz: Good catch.

src/main.rs
keszybz: But that'd just return from the closure, no?

nabijaczleweli: Oh, I'd missed the closure part; IIRC the log is explicitly flushed, so even though "no destructors on the current stack or any other thread's stack will be run", an `info!(); exit(0)` should be fine, if imperfect, but all alternatives I came up with are even uglier, so.

keszybz:
Pfff, time flies. Updated with a few minor typo fixes and `{info!("Disabled by kernel cmdline option"); return Ok(())}` as suggested.

keszybz:
... and rebased.

nabijaczleweli (approved):

PR #49, opened 2020-10-02
make: split systemd_service target out of build

Keruspe:
And also add generated unit to gitignore

Handy for packaging, when the cargo stuff is already handled outside of the Makefile


Keruspe:
Rebased

keszybz:
LGTM.

PR #52, opened 2021-01-03
Replace libc::getpid with std::process::id

vilgotf:
`std::process::id()` was added in rust 1.26.

I'm a bit unsure on the formatting, is it preferable to `use std::process` and then `process::id()` or is that wrong cause it imports unnecessary code into the namespace?


nabijaczleweli (approved):

vilgotf:
(Fixed typo in commit msg)

PR #55, opened 2021-01-12
Mount points

keszybz:


src/config.rs
```diff
@@ -13,11 +13,16 @@ use std::path::{Path, PathBuf};
...
+
     pub host_memory_limit_mb: Option<u64>,
+
     pub zram_fraction: f64,
     pub max_zram_size_mb: Option<u64>,
     pub compression_algorithm: Option<String>,
     pub disksize: u64,
+
+    pub mount_point: Option<PathBuf>, // only for mounts
+    pub fs_type: Option<String>,      // only for mounts
```
nabijaczleweli, line 25: The comment's not quite true, though, is it?

src/config.rs
```diff
@@ -204,6 +226,22 @@ fn parse_optional_size(val: &str) -> Result<Option<u64>> {
     })
 }
 
+fn verify_mount_point(val: &str) -> Result<PathBuf> {
+    let path = PathBuf::from(val);
```
nabijaczleweli, line 230: This likely wants to be `Path::new(val)`, since that won't allocate.

src/config.rs
```diff
@@ -204,6 +226,22 @@ fn parse_optional_size(val: &str) -> Result<Option<u64>> {
...
+
+    if path.is_relative() {
+        return Err(anyhow!("mount-point {} is not absolute", val));
+    }
+
+    for component in path.components() {
+        if let std::path::Component::ParentDir = component {
```
nabijaczleweli, line 237: No other bit uses fully-qualified identifiers, a `use` clause (potentially with a rename` would be more consistent.

Also, isn't this whole thing just `path.components().any(|c| c == Component::ParentDir)`?

src/generator.rs
```diff
@@ -90,45 +90,161 @@ pub fn run_generator(devices: &[Device], output_directory: &Path, fake_mode: boo
...
+    /* dev-zramX.swap */
 
+    write_contents(
+        output_directory,
+        &swap_name,
+        &format!(
+            "\
 [Unit]
-Description=Compressed swap on /dev/{zram_device}
+Description=Compressed Swap on /dev/{zram_device}
```
nabijaczleweli, line 158: Why the capitalisation change?

src/generator.rs
```diff
@@ -90,45 +90,161 @@ pub fn run_generator(devices: &[Device], output_directory: &Path, fake_mode: boo
...
+fn write_contents(output_directory: &Path, filename: &str, contents: &str) -> Result<()> {
+    let path = output_directory.join(filename);
+    make_parent(&path)?;
+
+    let contents = format!(
+        "\
+# Automatically generated by {exe_name:?}
+
+{contents}",
+        exe_name = std::env::current_exe().unwrap(),
```
nabijaczleweli, line 102: Likewise regarding the full path here.

Also, I think doing `.display()` here and removing the `:?` above would look less weird in the general case.

src/generator.rs
```diff
@@ -90,45 +90,161 @@ pub fn run_generator(devices: &[Device], output_directory: &Path, fake_mode: boo
...
+    /* systemd-zram-setup@.service.
+     * We use the packaged unit, and only need to provide a small drop-in. */
+
+    write_contents(
+        output_directory,
+        &format!(
+            "systemd-zram-setup@{}.service.d/bindsto-swap.conf",
+            device.name
+        ),
+        &format!(
```
nabijaczleweli, line 142: This is a static string.

nabijaczleweli (changes requested):

src/generator.rs
keszybz: Because the string is used as a title by systemd, non-capitalized looks strange in "Starting ...", "Stopping ...". I added an explanation to the commit message too.

src/config.rs
keszybz: Ack.

src/config.rs
keszybz: `::new(val)` fails with `expected 0 arguments`.

keszybz:
Thank you for the review. Updated.

src/config.rs
keszybz: Hmm, this compiles
```diff
diff --git src/config.rs src/config.rs
index 135ced8001..84a0679d11 100644
--- src/config.rs
+++ src/config.rs
@@ -228,7 +228,7 @@ fn parse_optional_size(val: &str) -> Result<Option<u64>> {
 }
 
 fn verify_mount_point(val: &str) -> Result<PathBuf> {
-    let path = PathBuf::from(val);
+    let path = Path::new(val);
 
     if path.is_relative() {
         return Err(anyhow!("mount-point {} is not absolute", val));
@@ -238,7 +238,7 @@ fn verify_mount_point(val: &str) -> Result<PathBuf> {
         return Err(anyhow!("mount-point {:#?} is not normalized", path));
     }
 
-    Ok(path)
+    Ok(path.to_path_buf())
 }
 
```
I'm not sure if that's better.

nabijaczleweli: Well, I'm unsure on why you're actually returning it from here, but I did miss that you are. In this case, then, yeah, the `PathBuf::from()` variant is better.

nabijaczleweli (approved):

Issue #53, opened 2021-01-10
Unable to start swap-create@ service with zstd compressor


ElXreno:
Could you create a new release? There is a commit in the master branch that fixed the zram creation with the zstd compressor.
In version 0.2.0 you cannot create zram with zstd, the error is as follows:
```bash
Jan 10 13:34:34 laptop.local zram-generator[64345]: Error: Failed to configure disk size into /sys/block/zram0/disksize
Jan 10 13:34:34 laptop.local zram-generator[64345]: Caused by:
Jan 10 13:34:34 laptop.local zram-generator[64345]:     Cannot allocate memory (os error 12)
```

Additional info:
**System:** Fedora 33 (KDE)
**SystemD version:** 246.7-2.fc33
**zram-generator version:** 0.2.0-4.fc33

ignatenkobrain:
Just curious, which commit?

ElXreno:
Apparently, there is no fix in the master branch, it's just the configurator works from my console, but the service refuses to start.
Sorry about the mistake. I will rename this issue.

Could you check if this problem is reproduced on your system @ignatenkobrain?

polarathene:
> it's just the configurator works from my console, but the service refuses to start.

I assume you checked the closed issue on zstd with same error message, but if not, you might want to look at [this](https://github.com/systemd/zram-generator/issues/50).

ElXreno:
Thanks for the reply! `modprobe -r zram && modprobe zram` didn't work for me, but `modprobe zstd` did!
Maybe we should detect this error, and show a possible solution instead of displaying `Cannot allocate memory`?

ElXreno:
This is a very interesting problem, because if you manually run `/usr/lib/systemd/system-generators/zram-generator`, everything works without problems and without `modprobe zstd`.

keszybz:
Do you have selinux enabled? Any avcs?

keszybz:
Hmm, in my tests, the zstd module is automatically loaded when the device with zstd compression is initialized. We *could* execute `modprobe <compression>` for any compression type that is used, but I'd like to understand why it is failing before doing that.

ElXreno:
> Do you have selinux enabled?

Of course. In Fedora it enabled by default.

polarathene:
Not an issue on new Ubuntu 20.10 VPS instance:

```
cat /proc/modules | grep -c zstd
# 0

apt-get update && apt install --yes build-essential pkg-config ronn
[...]

PR #56, opened 2021-01-12
Minor code refactoring

keszybz:


src/generator.rs
```diff
@@ -58,21 +57,24 @@ pub fn run_generator(devices: &[Device], output_directory: &Path, fake_mode: boo
...
             Command::new("modprobe")
                 .arg("zram")
                 .status()
                 .context("modprobe call failed")?;
         }
 
-        let max_device = devices_made.into_iter().fold(0, cmp::max);
+        let max_device = devices
+            .iter()
+            .map(|device| device.name[4..].parse().unwrap())
```
nabijaczleweli, line 79: This wants to host the `.with_context(|| format!("zram device \"{}\" number", device.name))` from `handle_device()`, AFAICT nothing prevents this same panic here as vs there.

src/generator.rs
```diff
@@ -4,6 +4,7 @@ use crate::config::Device;
 use anyhow::{anyhow, Context, Result};
 use log::{info, warn};
 use std::cmp;
+use std::collections::HashSet;
```
nabijaczleweli, line 7: Prefer BTreeSet, especially at this size.

src/generator.rs
```diff
@@ -90,6 +91,31 @@ pub fn run_generator(devices: &[Device], output_directory: &Path, fake_mode: boo
...
             {}
         }
+
+        let compressions: Vec<&String> = devices
+            .iter()
+            .filter(|device| !device.compression_algorithm.is_none())
+            .map(|device| device.compression_algorithm.as_ref().unwrap())
+            .collect::<HashSet<_>>()
+            .into_iter()
+            .collect();
```
nabijaczleweli, line 101: No need for this, just leave it as `BTreeSet<_>`.

src/generator.rs
```diff
@@ -90,6 +91,31 @@ pub fn run_generator(devices: &[Device], output_directory: &Path, fake_mode: boo
...
+
+        if !compressions.is_empty() {
+            let known = get_known_compressions(None)?;
+
+            for compression in compressions.into_iter().filter(|c| !known.contains(c)) {
```
nabijaczleweli, line 106: If you make `get_known_compressions()` return a `BTreeSet` (since, well, that's what it represents), you'll be able to do `compressions.intersection(known)` here, which won't be quadratic.

src/generator.rs
```diff
@@ -90,6 +91,31 @@ pub fn run_generator(devices: &[Device], output_directory: &Path, fake_mode: boo
                 < max_device
             {}
         }
+
+        let compressions: Vec<&String> = devices
+            .iter()
+            .filter(|device| !device.compression_algorithm.is_none())
+            .map(|device| device.compression_algorithm.as_ref().unwrap())
```
nabijaczleweli, line 98: This is just `.flat_map(|d| d.compression_algorithm.as_ref())`.

nabijaczleweli (changes requested):

src/generator.rs
keszybz: Hmm, but we already parsed it in `read_devices()`. So if it didn't fail there, it cannot fail here, so a panic is "good enough"?

src/generator.rs
keszybz: Duh.

keszybz:
So... based on the conversation in #53, the last commit doesn't help with anything, so I'll drop it.
This still leaves two first cleanup commits, which can go in. I'll force-push to remove the last commit.

src/generator.rs
nabijaczleweli: Ah yeah, forgot about that bit. In that case it's customary to do `.expect("already verified in read_devices()")` or something along those lines, for the benefit of both the casual reader and end-user.

keszybz: Updated like that.

PR #57, opened 2021-01-13
Allow swap priority to be configured

### LINKED ISSUE #8: RFE: should set priority
The generator should set a priority, otherwise in the case of more than one swap (conventional and zram), it's ambiguous which one is used first, since it's kernel assigned.

```
$ swapon
NAME       TYPE      SIZE USED PRIO
/dev/sda6  partition   8G   0B   -2
/dev/zram0 partition 7.7G   0B   -3
```
Priority can be set to 0 - 32767. I think zram-generator should default to something decently high like 30000. I'm not opposed to it being configurable.

keszybz:
Closes #8.

### DIFF
--- a/man/zram-generator.conf.md
+++ b/man/zram-generator.conf.md
@@ -80,6 +80,12 @@ Devices with the final size of *0* will be discarded.
 
   If unset, none will be configured and the kernel's default will be used.
 
+* `swap-priority`=
+
+  Controls the relative swap priority, a value between -1 and 32767. Higher numbers indicate higher priority.
+
+  If unset, 100 is used.
+
 * `mount-point`=
 
   Format the device with a file system (not as swap) and mount this file system over the specified directory.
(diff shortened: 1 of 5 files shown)


man/zram-generator.conf.md
```diff
@@ -80,6 +80,12 @@ Devices with the final size of *0* will be discarded.
 
   If unset, none will be configured and the kernel's default will be used.
 
+* `swap-priority`=
+
+  Controls the relative swap priority, a value between -1 and 32767. Higher numbers indicate higher priority.
```
polarathene, line 85: Isn't the range full i32?

keszybz: I took the range from swapon(8). I think the negative values are not supposed to be configured directly, but are used by the kernel.

keszybz:
OK, this is straightforward enough. I'll merge it without further review once the tests pass.

Issue #50, opened 2020-10-09
Problem enabling zstd


darkdragon-001:
I tried to use `zstd` compression with zram on Fedora 33 with kernel `5.8.14-300.fc33.x86_64`.

    # /etc/systemd/zram-generator.conf
    [zram0]
    compression-algorithm = zstd

Restarting via `sudo systemctl restart swap-create@zram0.service` gives me the following error:
```
Okt 09 16:12:03 localhost.localdomain systemd[1]: Starting Create swap on /dev/zram0...
Okt 09 16:12:03 localhost.localdomain kernel: Can't allocate a compression stream
Okt 09 16:12:03 localhost.localdomain kernel: zram: Cannot initialise zstd compressing backend
Okt 09 16:12:03 localhost.localdomain zram-generator[3938]: Error: Failed to configure disk size into /sys/block/zram0/disksize
Okt 09 16:12:03 localhost.localdomain zram-generator[3938]: Caused by:
Okt 09 16:12:03 localhost.localdomain zram-generator[3938]:     Cannot allocate memory (os error 12)
Okt 09 16:12:03 localhost.localdomain systemd[1]: swap-create@zram0.service: Main process exited, code=exited, status=1/FAILURE
```
I also tried rebooting and the same problem occurs.

```
$ cat /sys/block/zram0/comp_algorithm
lzo [lzo-rle] lz4 lz4hc 842 zstd
```

What is going on? Any help appreciated!

nabijaczleweli:
Does this also happen when you run this as root?
```sh
modprobe -r zram
modprobe zram
echo 1 > /sys/block/zram0/reset
echo zstd > /sys/block/zram0/comp_algorithm
echo $NUMBER > /sys/block/zram0/disksize
/usr/lib/systemd/systemd-makefs swap /dev/zram0
```
(Substitute `$NUMBER` for your desired disksize, of course.)

darkdragon-001:
@nabijaczleweli Thanks for your help! `modprobe zram` solved the problem (which I persisted across reboots via `/etc/modules-load.d/`). I was just confused because I use btrfs with zstd compression for my root filesystem (from which the systemd config files are loaded) - obviously this doesn't need the additional kernel module...

keszybz:
[...]

PR #60, opened 2021-01-14
fix: Load kernel module for compressors when required

polarathene:
While some distros like Ubuntu 20.10 are found to automatically load `zstd` or related modules when required for `zram-generator` to work, others like Fedora 33 do not mimic this behaviour. I am not sure what prevents or enables that difference in distro configuration, but we can detect if the compressor module is not loaded and attempt to load within `zram-generator`.

The fix queries `/proc/crypto` and parses it to build a list of enabled kernel crypto modules, some of which are of the `compression` type which `zram` uses. The `compression-algorithm` values from config file are collected and checked if they are present in this list, if not, `modprobe <compressor>` is called to load the relevant module.

---

A commit in [this PR](https://github.com/systemd/zram-generator/pull/56#issuecomment-759362184) was attempted earlier but later reverted as it was not successful. This PR includes code from @keszybz [reverted commit](https://github.com/keszybz/zram-generator/commit/2297c66207132f891af12df036ae5bce2dc5a8df#diff-01a1d809f0e7b93e591f14b055c3acc76e489ec56173eaead04003bb599bde3d) and adds my addition of `get_known_compressors()`.

---

This has been tested against a Fedora 33 instance (removed the `zram-generator` package and installed from source of my git branch after a reboot).  It works with the given caveats:

- `systemctl daemon-reload` will trigger the relevant `run_generators()` method, or `/path/to/zram-generator <some_directory>` will likewise trigger the method.
- `systemctl restart swap.target` or `systemctl start systemd-zram-generator@zram0.service` does not and will continue to result in failure. Thus if a user modifies their config in a running session and attempts to restart the relevant service, it's unclear how they can avoid the error without knowing to run `systemctl daemon-reload` or restarting their system.
- Ideally the proper way to handle this refresh/re-initialization/update is to **add documentation for how users should apply config updates.**

---

Fixes: #50, Fixes: #53

### DIFF
--- a/src/generator.rs
+++ b/src/generator.rs
@@ -4,6 +4,7 @@ use crate::config::Device;
 use anyhow::{anyhow, Context, Result};
 use log::{info, warn};
 use std::cmp;
+use std::collections::BTreeSet;
 use std::fs;
 use std::os::unix::fs::symlink;
 use std::path::Path;
@@ -96,7 +97,43 @@ pub fn run_generator(devices: &[Device], output_directory: &Path, fake_mode: boo
         }
     }
 
+    let compressors: BTreeSet<&String> = devices
+        .iter()
+        .flat_map(|device| device.compression_algorithm.as_ref())
+        .collect::<BTreeSet<_>>();
+
+    if !compressors.is_empty() {
+        let known = get_known_compressors()?;
+
+        for comp in compressors.into_iter().filter(|&c| !known.contains(c)) {
+            let status = Command::new("modprobe")
+                .arg(format!("crypto-{}", comp))
+                .status()
+                .context("Failed to spawn modprobe")?;
+            if !status.success() {
+                warn!("modprobe crypto-{} failed, ignoring: code {}", comp, status);
+            }
+        }
+    }
+
     Ok(())
+}
+
+// Returns a list of kernel crypto modules, including available (loaded) compressors
+fn get_known_compressors() -> Result<BTreeSet<String>> {
+    let path = Path::new("/proc/crypto");
+    let content = fs::read_to_string(path).with_context(|| format!("Failed to read {:?}", path))?;
+
+    // Extract algorithm names (this includes non-compression algorithms too)
+    let available = content
+        .lines()
+        .into_iter()
+        .filter(|line| line.starts_with("name"))
+        .map(|m| m.rsplit(':').next().unwrap().trim())
+        .map(String::from)
+        .collect();
+
+    Ok(available)
 }
 
 fn write_contents(output_directory: &Path, filename: &str, contents: &str) -> Result<()> {
--- a/src/generator.rs
+++ b/src/generator.rs
@@ -97,15 +97,17 @@ pub fn run_generator(devices: &[Device], output_directory: &Path, fake_mode: boo
         }
     }
 
-    let compressors: BTreeSet<&String> = devices
+    let compressors: BTreeSet<_> = devices
         .iter()
-        .flat_map(|device| device.compression_algorithm.as_ref())
-        .collect::<BTreeSet<_>>();
+        .flat_map(|device| device.compression_algorithm.as_deref())
+        .collect();
 
     if !compressors.is_empty() {
-        let known = get_known_compressors()?;
+        let proc_crypto =
+            fs::read_to_string("/proc/crypto").context("Failed to read /proc/crypto")?;
+        let known = parse_known_compressors(&proc_crypto);
 
-        for comp in compressors.into_iter().filter(|&c| !known.contains(c)) {
+        for comp in compressors.difference(&known) {
             let status = Command::new("modprobe")
                 .arg(format!("crypto-{}", comp))
                 .status()
@@ -119,25 +121,15 @@ pub fn run_generator(devices: &[Device], output_directory: &Path, fake_mode: boo
     Ok(())
 }
 
-fn _get_known_compressors(path: &Path) -> Result<BTreeSet<String>> {
-    let content = fs::read_to_string(path).with_context(|| format!("Failed to read {:?}", path))?;
-
+// Returns a list of names of loaded compressors
+fn parse_known_compressors(proc_crypto: &str) -> BTreeSet<&str> {
     // Extract algorithm names (this includes non-compression algorithms too)
-    let available = content
+    proc_crypto
         .lines()
         .into_iter()
         .filter(|line| line.starts_with("name"))
         .map(|m| m.rsplit(':').next().unwrap().trim())
-        .map(String::from)
-        .collect();
-
-    Ok(available)
-}
-
-// Returns a list of names of loaded compressors
-fn get_known_compressors() -> Result<BTreeSet<String>> {
-    let path = Path::new("/proc/crypto");
-    _get_known_compressors(&path)
+        .collect()
 }
 
 fn write_contents(output_directory: &Path, filename: &str, contents: &str) -> Result<()> {
@@ -295,15 +287,11 @@ Where={mount_point:?}
 #[cfg(test)]
 mod tests {
     use super::*;
-    use std::io::Write;
     use std::iter::FromIterator;
 
     #[test]
-    fn test_get_known_compressors() {
-        let mut file = tempfile::NamedTempFile::new().unwrap();
-
-        file.write(
-            b"\
+    fn test_parse_known_compressors() {
+        let data = "\
 name         : zstd
 driver       : zstd-scomp
 module       : zstd
@@ -341,18 +329,8 @@ refcnt       : 2
 selftest     : passed
 internal     : no
 type         : skcipher
-",
-        )
-        .unwrap();
-        file.flush().unwrap();
-        let expected = vec![
-            String::from("zstd"),
-            String::from("ccm(aes)"),
-            String::from("ctr(aes)"),
-        ];
-        assert_eq!(
-            _get_known_compressors(file.path()).unwrap(),
-            BTreeSet::from_iter(expected)
-        );
+";
+        let expected = vec!["zstd", "ccm(aes)", "ctr(aes)"];
+        assert_eq!(parse_known_compressors(data), BTreeSet::from_iter(expected));
     }
 }
--- a/src/generator.rs
+++ b/src/generator.rs
@@ -119,9 +119,7 @@ pub fn run_generator(devices: &[Device], output_directory: &Path, fake_mode: boo
     Ok(())
 }
 
-// Returns a list of kernel crypto modules, including available (loaded) compressors
-fn get_known_compressors() -> Result<BTreeSet<String>> {
-    let path = Path::new("/proc/crypto");
+fn _get_known_compressors(path: &Path) -> Result<BTreeSet<String>> {
     let content = fs::read_to_string(path).with_context(|| format!("Failed to read {:?}", path))?;
 
     // Extract algorithm names (this includes non-compression algorithms too)
@@ -134,6 +132,12 @@ fn get_known_compressors() -> Result<BTreeSet<String>> {
         .collect();
 
     Ok(available)
+}
+
+// Returns a list of names of loaded compressors
+fn get_known_compressors() -> Result<BTreeSet<String>> {
+    let path = Path::new("/proc/crypto");
+    _get_known_compressors(&path)
 }
 
 fn write_contents(output_directory: &Path, filename: &str, contents: &str) -> Result<()> {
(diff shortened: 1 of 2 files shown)


polarathene:
## Reproduction steps

### Ubuntu 20.10 (via Vultr):

```
apt-get update && apt install --yes build-essential pkg-config ronn
curl https://sh.rustup.rs -sSf | sh -s -- -y
source $HOME/.cargo/env
git clone https://github.com/systemd/zram-generator
cd zram-generator && make build && make install
echo -e "[zram0]\ncompression-algorithm = zstd\n[zram1]compression-algorithm = lz4hc\n[zram2]" >> /etc/systemd/zram-generator.conf
systemctl daemon-reload
```
`systemctl start swap.target` is required or a `shutdown -r now`, otherwise `zramctl` and `swapon` will not output anything, but `dmesg` will still log `zram-generator` creating swap units and `zram` adding devices. Running the `zstd` check (0 not loaded, >=1 loaded) prior to one of these two final commands will show `0`, thus Ubuntu is autoloading `zstd`.

#### Check for success

```
$ swapon
NAME       TYPE      SIZE USED PRIO
/dev/zram0 partition 488M   0B  100
/dev/zram1 partition 488M   0B  100
/dev/zram2 partition 488M   0B  100

$ zramctl
NAME       ALGORITHM DISKSIZE DATA COMPR TOTAL STREAMS MOUNTPOINT
/dev/zram2 lzo-rle       486M   4K   73B   12K       1 [SWAP]
/dev/zram1 lz4hc         486M   4K   62B    4K       1 [SWAP]
/dev/zram0 zstd          486M   4K   58B    4K       1 [SWAP]

$ cat /proc/modules | grep -c zstd
1
```

### Fedora 33 (via Vultr):

```
dnf remove --assumeyes zram-generator
dnf install --assumeyes @development-tools
curl https://sh.rustup.rs -sSf | sh -s -- -y
source $HOME/.cargo/env
git clone https://github.com/systemd/zram-generator
cd zram-generator && make build && make install
echo -e "[zram0]\ncompression-algorithm = zstd\n[zram1]compression-algorithm = lz4hc\n[zram2]" >> /etc/systemd/zram-generator.conf
shutdown -r now
```

Fedora uses `shutdown -r now` instead of `systemctl daemon-reload` to ensure a cleaner reference. Otherwise the existing services/swap remains active with the now removed `swap-create@` service files (which have now been replaced with `systemd-zram-setup@` services from recent changes in git).

#### Check for success

```
$ swapon
NAME       TYPE      SIZE USED PRIO
/dev/zram2 partition 486M   0B  100

$ zramctl
NAME       ALGORITHM DISKSIZE DATA COMPR TOTAL STREAMS MOUNTPOINT
/dev/zram2 lzo-rle       486M   4K   73B   12K       1 [SWAP]

$ cat /proc/modules | grep -c zstd
0
```

`zstd` and `lz4hc` failed to load. Modules aren't loaded and weren't loaded like with Ubuntu. Explicitly start the generated zram0 (`zstd`) service:

```
$ systemctl start systemd-zram-setup@zram0.service
Job for systemd-zram-setup@zram0.service failed because the control process exited with error code.
See "systemctl status systemd-zram-setup@zram0.service" and "journalctl -xe" for details.

$ journalctl
Jan 14 06:46:15 vultr.guest systemd[1]: Starting Create swap on /dev/zram0...
Jan 14 06:46:15 vultr.guest kernel: Can't allocate a compression stream
Jan 14 06:46:15 vultr.guest kernel: zram: Cannot initialise zstd compressing backend
Jan 14 06:46:15 vultr.guest zram-generator[813]: Error: Failed to configure disk size into /sys/block/zram0/disksize
Jan 14 06:46:15 vultr.guest zram-generator[813]: Caused by:
Jan 14 06:46:15 vultr.guest zram-generator[813]:     Cannot allocate memory (os error 12)
Jan 14 06:46:15 vultr.guest systemd[1]: systemd-zram-setup@zram0.service: Main process exited, code=exited, status=1/FAILURE
Jan 14 06:46:15 vultr.guest systemd[1]: systemd-zram-setup@zram0.service: Failed with result 'exit-code'.
Jan 14 06:46:15 vultr.guest systemd[1]: Failed to start Create swap on /dev/zram0.
Jan 14 06:46:15 vultr.guest systemd[1]: Dependency failed for Compressed Swap on /dev/zram0.
Jan 14 06:46:15 vultr.guest systemd[1]: dev-zram0.swap: Job dev-zram0.swap/start failed with result 'dependency'.
```

This is the usual expected behaviour and users/admins are generally expected to configure their systems to load kernel modules themselves rather than a utility handling it afaik. A reproduction can be done with this PR branch instead of upstream master replacing the git repo to:

```
git clone https://github.com/polarathene/zram-generator --single-branch --branch "fix/load_compression_modules"
```

You can do this with the existing reproduction if you like without having to restart, you'll need to run `systemctl daemon-reload && systemctl start swap.target`. The first command should load the kernel modules now, and the 2nd trigger the zram-generator generated services, which are now successful:

```
$ swapon
NAME       TYPE      SIZE USED PRIO
/dev/zram2 partition 486M 4.3M  100
/dev/zram1 partition 486M   0B  100
/dev/zram0 partition 486M   0B  100

$ zramctl
NAME       ALGORITHM DISKSIZE DATA COMPR TOTAL STREAMS MOUNTPOINT
/dev/zram2 lzo-rle       486M   4M    1M  1.6M       1 [SWAP]
/dev/zram1 lz4hc         486M   4K   62B    4K       1 [SWAP]
/dev/zram0 zstd          486M   4K   58B    4K       1 [SWAP]

$ cat /proc/modules | grep -c zstd
1
```

keszybz:
Thanks. I updated your PR a bit: I applied @nabijaczleweli's comments from my PR here and simplified the code a bit. 

Instead of checking the 'driver:' line, it seems better to check the 'name:' line, since that's what should actually match the name specified in `comp_algorithm` in zram configuration.

I also added a unittest for the new function and a proper commit message ;)

> While some distros like Ubuntu 20.10 are found to automatically load `zstd` or related modules when required for `zram-generator` to work, others like Fedora 33 do not mimic this behaviour.

As I said before, the distro is not that important here. Both distros use very similar kernels, and the fact that you're observing different behaviour is either because of different kernel versions or because of timing differences.

There should be no functional change, but if you could give this version a spin and test that it still works for you, that'd be great.

polarathene:
> Instead of checking the 'driver:' line, it seems better to check the 'name:' line, since that's what should actually match the name specified in `comp_algorithm` in zram configuration.

I avoided `name` because duplicates would need to be filtered out (not a problem since you switched to a set), `-generic` suffix drivers are the ones that are typed for compression use, it doesn't make a difference either way with the code, but was at least more explicit about which one was relevant.

Doesn't really matter about matching `comp_algorithm` as you still get that correct value out of it. We could modprobe those modules too, but for whatever reason you modprobe the alias `crypto-<module>`.

---

> I also added a unittest for the new function and a proper commit message ;)

I usually provide more detailed commit messages, but I had to use the CLI for this PR, of which I'm not familiar of a nice way to write extensive multi-line commit messages. Usually a PR on github provides context for the commit merged to master so I didn't see it as too much of a concern (assuming this PR is merged into a single commit or rebased to one with different commit message as is usually the case with my contributions).

This is inaccurate btw (or at least the way it's been phrased), I've tried to communicate that clearly to you already:

> *The kernel normally loads the compressor module when a device with a given compression module is loaded, but this doesn't always work.*

Same with this, that applies to any zram block device that's already been initialized, and is like other block devices providing choices such as I/O schedulers:

> *Note: /sys/block/zram0/comp_algorithm lists compressors known to zram, not the ones currently loaded.*

And the error message at the top of the commit message is not "occasionally" failing, it's due to the compressor module not being loaded and is consistent on a system where that failure is happening:

> *Creation of a zram device (with zstd as the compression alg) occasionally fails with ENOMEM:*

Cheers for the unit test! Rewriting the history seemed unnecessary however, which made it a bit difficult for me to spot the changes since my commit. That could be done when merging the PR.

---

> As I said before, the distro is not that important here. Both distros use very similar kernels, and the fact that you're observing different behaviour is either because of different kernel versions or because of timing differences.

As I said before, there's a clear reproducible and consistent difference in behaviour between the two. It is most certainly not a timing difference and I'd be very doubtful of it being some regression in a patch version.

**EDIT:** I've identified the cause, it's due to an SELinux policy. `setenforce 0` temporarily disables it and allows for desired/expected behaviour. I know you [considered SELinux earlier](https://github.com/systemd/zram-generator/issues/53#issuecomment-759327578) but dismissed it without advising to try disabling it, I've never dealt with SELinux prior so I had no clue what to do or if it could prevent the kernel module loading. My understanding is it's unrelated to `modprobe` call but some sort of implicit equivalent of loading a module when needed that the policy prevents.

So while kernel versions are similar, the difference is this security module, which I tried to raise as the kernels being configured/built differently as the differentiator..

---

Thanks for taking the time to review and help troubleshoot this, along with the added contributions to this PR.

src/generator.rs
```diff
@@ -96,9 +97,50 @@ pub fn run_generator(devices: &[Device], output_directory: &Path, fake_mode: boo
...
+        }
+    }
+
     Ok(())
 }
 
+fn _get_known_compressors(path: &Path) -> Result<BTreeSet<String>> {
+    let content = fs::read_to_string(path).with_context(|| format!("Failed to read {:?}", path))?;
+
+    // Modules with type of `compression` have a `-generic` suffix in the driver name
```
polarathene, line 126: This comment isn't relevant anymore with your change from `driver` to `name` matching? Or keeping it for some context to anyone visiting this code and wondering what's going on? (I noticed my other comment for the now split function was rephrased to exclude mention that not all results would be compressor names)

```suggestion
    // Extracts the `name` field values from `/proc/crypto` output, some of which are compressor module names
```

src/generator.rs
```diff
@@ -96,9 +97,50 @@ pub fn run_generator(devices: &[Device], output_directory: &Path, fake_mode: boo
         }
     }
 
+    let compressors: BTreeSet<&String> = devices
+        .iter()
+        .filter(|device| device.compression_algorithm.is_some())
+        .flat_map(|device| device.compression_algorithm.as_ref())
```
nabijaczleweli, line 103: Having these two one after another is redundant: `Option::into_iter()` yields no elements for `None` and one element for `Some`, so filtering out `None` variants beforehand does nothing.

src/generator.rs
```diff
@@ -96,9 +97,50 @@ pub fn run_generator(devices: &[Device], output_directory: &Path, fake_mode: boo
...
+        .collect::<BTreeSet<_>>();
+
+    if !compressors.is_empty() {
+        let known = get_known_compressors()?;
+
+        for comp in compressors.into_iter().filter(|&c| !known.contains(c)) {
```
nabijaczleweli, line 109: This is `compressors.difference(&known)`. Granted, that will not work out of box since one is `BTreeSet<&String>` and the other is `BTreeSet<String>`.

If I were to go full shitlord on this, I'd lift the `fs::read_to_string()` to be parallel to `known`, pass it to `get_known_compressors()` (now known as `parse_known_compressors()` or sth), and make the compressors flatmap be over `d.c_a.as_ref().map(String::as_str)` (or `d.c_a.as_ref().map(|a| &a[..])` as the case may be) to have them both be `BTreeSet<&str>` (and save on some needless allocations) to spell this out correctly semantically.

Especially since the `get_known_compressors()`/`_get_known_compressors()` looks like it does now, as opposed to how it did last time around.

src/generator.rs
```diff
@@ -96,9 +97,50 @@ pub fn run_generator(devices: &[Device], output_directory: &Path, fake_mode: boo
...
+        .into_iter()
+        .filter(|line| line.starts_with("name"))
+        .map(|m| m.rsplit(':').next().unwrap().trim())
+        .map(String::from)
+        .collect();
+
+    Ok(available)
+}
+
+// Returns a list of names of loaded compressors
```
polarathene, line 138:
```suggestion
// Returns a list of names that includes loaded modules of compressors
```

nabijaczleweli (changes requested):
The second one is less of a hard Change Request, but the first one very much is.

polarathene (commented):
For the benefit of others coming across the code, or future selfs, a little more clarity in these two comments can go a long way :)

src/generator.rs
keszybz: Actually, those are the names of *compressors*, not *module* names. (A module may implement multiple algorithms.) And we don't particularly care about modules themselves.

keszybz:
Updated again. I removed the unnecessary `filter` that @nabijaczleweli pointed out, and the stale comment.
I also added more stuff to the tests that was omitted before.

I didn't do the `compressors.difference(&known)` change. I'm not sure if I know how... @nabijaczleweli please submit a separate patch if you think it's worth it ;)

nabijaczleweli:
Implementation in https://github.com/nabijaczleweli/systemd-zram-generator/commit/4aa8efdf5a9db15c61d6a20ddba21e523e637fc9, straight-forward as it gets, ff against this branch. Pull it in here before merging if you want.

nabijaczleweli (approved):

keszybz:
Thanks. I pulled in your patch. clippy said to use `device.compression_algorithm.as_deref()`, which makes the whole thing even a bit shorter. I also added another patch to only warn if we cannot read the file.

daiaji:
Has this bug not been fixed? Zram generator 1.1.1 still seems to have this problem, or is it a kernel error? But manjaro's kernel seems to turn off SELinux by default.

nabijaczleweli:
It has, and this fix is still in: https://github.com/systemd/zram-generator/blob/20b9261790266be3f9cbb1a42e0c92ce581b9587/src/generator.rs#L118-L133

Is there any particular module that's not being loaded for you?

daiaji:
I use BTRFS and I use ZSTD compression and I think I have the ZSTD module loaded.

daiaji:
Should I wait to fix the patch?

nabijaczleweli:
The fix is to stop using whatever's providing udisks2-zram-setup@

daiaji:
It seems that the udev left by the zram that I manually set up according to Archwiki has not been deleted. It seems to cause a conflict with zram-generator. After deleting the residual files,
zram-generator is working properly.

PR #63, opened 2021-01-22
generator: do not fail if modprobe is not found

keszybz:
The tests failed in Fedora mock build, without modprobe installed.
We don't fail if the modprobe call fails, so we also shouldn't if the
command is unavailable.


nabijaczleweli (approved):

keszybz:
Thank you for the review.

PR #64, opened 2021-01-22
Specify disk size as an arbitrary mathematical expression

### LINKED ISSUE #51: RFE: support boosting zram-fraction on low memory machines
Hi,

As you may know I've done (and still do) a lot of work to make sure that Linux and specifically Fedora works well on Intel Bay Trail based tablets and 2-in-1s.

Some of these come with only 1G of RAM, having ZRAM enabled by default is great for these, but the fraction of 0.5 is too low. Too be able to install the Fedora 33 workstation livecd on one of these devices I need to swapoff, reconfigure zram0 to be 1G (so fraction is 1.0) and then mkswap + swapon again. Otherwise anaconda won't run.

I realize that using fraction 1.0 (or more) is a bit controversial, but generally speaking the zram compression is equal to or better then 1:3, so this takes about 1/3th of the RAM leaving the rest for normal usage.

I've been considering doing a Fedora Workstation "light" spin for these devices where changing the zram fraction would be one of the tweaks, but it would be better if we can just make things work with the main Fedora Workstation release.

As such I was wondering if it would be possible to do something for these devices, maybe add support for something like this to the config file:

```
# Settings for machines with a low amount of RAM. Machines with less RAM then the low-memory-limit
# will use the optional zram-fraction-low-mem setting as zram-fraction. This allows using a higher
# zram-fraction on low memory machines, which need all the extra (z)RAM they can get.
# The default value for both is 0, which disables this feature.
low-memory-limit = 1536
zram-fraction-low-mem = 1.0
```

Note if this is considered something which you would be willing to add when contributed, please also let me know then I will try to add this (I have 0 experience coding in rust though...).

keszybz:
Closes #51.

It's like using a cannon on a mosquito. Also, such expressions are not
very user friendly. OTOH, it allows rather complex policies to be
implemented.

Debug build size grows from 17MB to 20MB, so not *too* bad.

I implemented this mostly to see how it would look. I'm not convinced the final effect is worth all the complexity.


jwrdegoede:
Interesting, not sure if this is necessary with the decision to just go with a fraction of 1.0 in Fedora, but definitely interesting and better then my proposal from the RFE, as this is much more flexible.

src/config.rs
```diff
@@ -364,6 +381,17 @@ pub fn kernel_zram_option(root: &Path) -> Option<bool> {
...
 }
 
+pub fn eval_size_expression(expr: &str, ram_size: u64) -> Result<f64> {
+    let x = Expr::new(expr)
+        .value("ram", ram_size)
+        .exec()
+        .with_context(|| format!("Expression \"{}\" cannot be evaluated", expr))?;
+    match x.as_f64() {
+        Some(x) => Ok(x),
+        None => Err(anyhow!("Expression evaluated to \"{}\", which is not a number", x))
```
nabijaczleweli, line 391: `x.as_f64().or_else(|x| anyhow!("", x))`

nabijaczleweli (changes requested):

src/config.rs
keszybz: That doesn't seem to work. What about this:
```rust
x.as_f64().
        with_context(|| format!("{}: expression evaluated to \"{}\", which is not a number", key, x))
```

keszybz:
Let's leave this open. I don't think the justification is strong enough now. But maybe people will find other uses (or a better way to do this...).

nabijaczleweli (approved):

ignatenkobrain (approved):
This looks cool :)

keszybz:
Replaced by #121.

PR #67, opened 2021-01-30
Use Box<str> when resize is unnecessary

vilgotf:
According to https://users.rust-lang.org/t/use-case-for-box-str-and-string/8295/4 `Box<str>` seems more efficient than `String`. I also think that it makes the intent clearer, use String when we need to be able to modify something, use str when we don't.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -17,9 +17,9 @@ enum Opts {
     /// Generate units into the directory
     GenerateUnits(String),
     /// Set up a single device
-    SetupDevice(String),
+    SetupDevice(Box<str>),
     /// Reset (destroy) a device
-    ResetDevice(String),
+    ResetDevice(Box<str>),
 }
 
 fn get_opts() -> Opts {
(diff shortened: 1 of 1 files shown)


nabijaczleweli (changes requested):
(a) Have you verified that these actually live long enough (i.e. the variants are `Opts::SetupDevice(Cow::Borrowed(..))`)? I doubt this.
(b) If so, just make these all `&str`s.

vilgotf:
Hmm, I'm fairly new to Rust so I don't really understand what you are asking of me.

nabijaczleweli:
I think it's quite clear – check whether the inner Cows are actually Borrowed at point of use, and if they are, make them (and the remaining String) plain &strs.

vilgotf:
But there are no Cows?

nabijaczleweli:
Wait, no, you're right, this is true. I just saw `Nnn<str>` and *hard* overfit to `Cow<str>`, heh.

But this only furthers that there's little use to this. As you've noted, `Opts` already contains a `String`, and it must continue to do so, and there's exactly zero (0) benefit to using a `Box<str>` (being an usize smaller isn't one, especially since `Opts` remains 32-bytes long) in this case, which outweighs making the variants non-uniform (and having to deal with a `Box<str>` in future) and appear as if they're different semantically (they aren't).

So, uh, thanks for the PR, but I will close this, as it improves nothing, and adds undue complexity.

vilgotf:
Yeah that's fine, as I said, I was mostly curious / looking for another opinion on `Box<str>`. Would you say that using `Box<str>` is preferable if all returns were `Box<str>`?

nabijaczleweli:
Naaah, `Box<str>` is a speciality type to be pulled out only when you Definitely need it, and uh, this really doesn't, and you lose out on (a) ergonomics of String and (b) not having to deal with `Box`es. Your link with rustc's interner is a case where they've clearly measured it and they definitely *do* need it, and it's a tangible go-fast-and-use-less-memory measure.

PR #79, opened 2021-06-08
remove utf-8 env validator

vilgotf:
This makes the code prettier but it could be seen as a regression.


vilgotf:
Ci failure is unrelated

nabijaczleweli (approved):

PR #83, opened 2021-07-02
make: rework targets

### LINKED ISSUE #82: Build error when doing 'make install'
Hello, after compiling from this repository, it displays following error and requires a fix:
install: cannot stat 'man/zram-generator.8': File or directory does not exist

Compile log can be found here:
https://gist.github.com/AhRoon/7c2aaea3f87b04970333d8d62756755d

I'm using Solus 4.2 "Fortitude" and latest version (Since it is rolling release I can't tell the exact version, sorry)

keszybz:
'make && make install' did not work because the man page wasn't listed
in dependencies. Let's redo the whole scheme:
'make build' builds everything,
'make program', 'make man', 'make systemd-service' build each of
the three items, and 'make install' should now always work.

"systemd_service" is renamed to "systemd-service". Dashes are normally
used in Makefile target names, no need to bother with underscores.

Fixes #82.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,7 @@ jobs:
           override: true
           profile: minimal
       - name: Build
-        run: make build CARGOFLAGS="--verbose"
+        run: make program CARGOFLAGS="--verbose"
       - name: Run tests
         run: make check CARGOFLAGS="--verbose"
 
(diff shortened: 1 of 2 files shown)


keszybz:
The two failures are from "module zram not found". We should probably skip those tests then...

hswong3i:
> The two failures are from "module zram not found". We should probably skip those tests then...

The broken `modprobe zram` issue had been addressed at #84 ;-)

keszybz:
Rebased.

PR #85, opened 2021-07-13
Makefile: Rewrok Targets

### LINKED ISSUE #82: Build error when doing 'make install'

hswong3i:
- Rewrite `Makefile` so we could manage it with `make && make test && sudo make install`
  - Split `make man` and `make install.man`, therefore could simply skip when packaging with RPM/DEB/etc
  - Add `make vendor` as wrapper of `cargo vendor`, therefore support `cargo build --offline` when running with build.opensuse.org

Fixes #82
Fixes #83

### DIFF
--- a/src/generator.rs
+++ b/src/generator.rs
@@ -284,10 +284,10 @@ After=systemd-zram-setup@{zram_device}.service
 
 [Mount]
 What=/dev/{zram_device}
-Where={mount_point:?}
+Where={mount_point}
 ",
             zram_device = device.name,
-            mount_point = device.mount_point.as_ref().unwrap(),
+            mount_point = device.mount_point.as_ref().unwrap().to_str().unwrap(),
         ),
     )?;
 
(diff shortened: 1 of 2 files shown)


keszybz:
>     * Rewrite `Makefile` so we could manage it with `make && make test && sudo make install`

That part would be OK.

>     * Split `make man` and `make install.man`, therefore could simply skip when packaging with RPM/DEB/etc

Is this really useful? When packaging, it's simple enough to `rm` the file or use `%exclude`. The installation targets don't have to support or possible details of packaging.

>     * Add `make vendor` as wrapper of `cargo vendor`, therefore support `cargo build --offline` when running with build.opensuse.org

I'm not a big fan of vendoring, but I would be OK with adding a Makefile wrapper for it. But it needs to be fully opt-in. Normal workflows must not use it.

PR #86, opened 2021-07-14
README.md: Rework with Installation Procedure

hswong3i:
- Improve `README.md` with installation and testing procedure
  - Re-format `README.md` with `pandoc -f gfm -t gfm -o README.md
    README.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,17 +1,20 @@
 # `systemd-zram-setup@.service` generator for zram devices
 
-This generator provides a simple and fast mechanism to configure swap on `/dev/zram*` devices.
+This generator provides a simple and fast mechanism to configure swap on
+`/dev/zram*` devices.
 
-The main use case is create **swap** devices, but devices with a file system can be created too, see below.
+The main use case is create **swap** devices, but devices with a file
+system can be created too, see below.
 
-### Configuration
+## Configuration
 
-A default config file may be located in /usr.
-This generator checks the following locations:
-* `/run/systemd/zram-generator.conf`
-* `/etc/systemd/zram-generator.conf`
-* `/usr/local/lib/systemd/zram-generator.conf`
-* `/usr/lib/systemd/zram-generator.conf`
+A default config file may be located in /usr. This generator checks the
+following locations:
+
+  - `/run/systemd/zram-generator.conf`
+  - `/etc/systemd/zram-generator.conf`
+  - `/usr/local/lib/systemd/zram-generator.conf`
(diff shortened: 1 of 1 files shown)


hswong3i:
@keszybz the key benefit for using pandoc is: it simply handle both code lint and syntax check for us, where the result is also predictable and repeatable.

README.md
```diff
@@ -20,70 +23,194 @@ In addition, "drop-ins" will be loaded from `.conf` files in
...
 It is recommended to use an existing package:
 
-* Fedora: `sudo dnf install zram-generator-defaults` (or `sudo dnf install zram-generator` to install without the default configuration)
-* Debian: packages provided by nabijaczleweli, see https://debian.nabijaczleweli.xyz/README.
-* Arch: `sudo pacman -S zram-generator` (or https://aur.archlinux.org/packages/zram-generator-git/ for the latest git commit)
+#### Fedora 33+
+
+``` bash
+sudo dnf config-manager --add-repo https://download.opensuse.org/repositories/home:alvistack/Fedora_33/home:alvistack.repo
+sudo dnf install zram-generator zram-generator-defaults
```
keszybz, line 74: There is no point in recommending personal packages when there's an official build in the distribution. We did that for Debian only because there was no official package. As soon as the official package is provided, the readme would be updated to point to that.

README.md
```diff
@@ -20,70 +23,194 @@ In addition, "drop-ins" will be loaded from `.conf` files in
...
+cd /etc/yum.repos.d/
+wget https://download.opensuse.org/repositories/home:alvistack/CentOS_8/home:alvistack.repo
+sudo yum install zram-generator zram-generator-defaults
+```
+
+#### Debian 10+
+
+``` bash
+echo 'deb http://download.opensuse.org/repositories/home:/alvistack/Debian_10/ /' | sudo tee /etc/apt/sources.list.d/home:alvistack.list
+curl -fsSL https://download.opensuse.org/repositories/home:alvistack/Debian_10/Release.key | gpg --dearmor | sudo tee /etc/apt/trusted.gpg.d/home_alvistack.gpg > /dev/null
```
keszybz, line 89: Why change from the repo provided by nabijaczleweli to this? Such a change would require justification.

keszybz (commented):
Hmm, so this changes the installation instructions from normal repository packages and the repo provided by one of the authors to some unknown source. Sorry, I have no idea who that is, and I wouldn't be comfortable recommending an unknown source even for those distributions where no other package exists.

PR #90, opened 2021-07-25
mount-options: Add Mount Options Support

### LINKED ISSUE #89: mount-options: Add Mount Options Support?
I am now using zram-generator to create `/var/lib/libvirt/images` for running my KVM for GitLab CI, e.g.
```
# zramctl 
NAME       ALGORITHM DISKSIZE  DATA COMPR TOTAL STREAMS MOUNTPOINT
/dev/zram1 lzo-rle     125.9G 37.6G 22.8G 23.3G      32 /var/lib/libvirt/images
/dev/zram0 lzo-rle         8G    4K   74B   12K      32 [SWAP]
```
After mount point created, I manually change the ownership and permission with:
```
# chown root:libvirt /var/lib/libvirt/images
# chmod 0775 /var/lib/libvirt/images
```

BTW, after every reboot, the `/var/lib/libvirt/images` ownership and permission reset as below:
```
# ls -ald /var/lib/libvirt/images
drwxr-xr-x 3 root root 4096 Jul 16 11:30 /var/lib/libvirt/images
```

Besides `mount-point` and `fs-type`, would that be possible to support `mount-options`, e.g. setting `mode=`, `uid=`, `gid=` (similar as tmpfs, see https://linux.die.net/man/8/mount)?

hswong3i:
When using zram for file system, we need to enable `discard` mount option for issue discard/TRIM commands to the underlying zram block device when blocks are freed. Else deleted files are not zero-ed, so zram does not remove the compressed pages.

Fixes #89


src/config.rs
```diff
@@ -38,6 +39,7 @@ impl Device {
             disksize: 0,
             swap_priority: 100,
             mount_point: None,
+            mount_options: Some("".to_string()),
```
nabijaczleweli, line 42: This makes no sense. Either make this just String, or actually only write it if it's set.

nabijaczleweli (changes requested):

src/config.rs
hswong3i: @nabijaczleweli sorry that I have zero experience on rust, and what I could handle is modify from similar logic, e.g. `fs-type` and `mount-point`.

What am I looking for from here is: if `mount-options` is empty string or None, just print an empty string to the generated file with `Options=""`.

May you provide some code snippet as reference, so I could modify it correctly?

hswong3i:
@nabijaczleweli test case added and passing as expected; BTW, after reboot the device is now format and override as swap by systemd, so the mount failed (because `discard` is not support):

```
# cat /etc/systemd/zram-generator.conf 
[zram0]
max-zram-size = 8192
zram-fraction = 1.0
[zram1]
max-zram-size = 8192
zram-fraction = 1.0
mount-point = /var/lib/libvirt/images
fs-type = ext4
mount-options = discard
```
```
# systemctl status systemd-zram-setup@zram1.service 
● systemd-zram-setup@zram1.service - Create swap on /dev/zram1
     Loaded: loaded (/lib/systemd/system/systemd-zram-setup@.service; static)
    Drop-In: /run/systemd/generator/systemd-zram-setup@zram1.service.d
             └─bindsto-mount.conf
     Active: inactive (dead) since Mon 2021-07-26 22:25:53 CST; 45s ago
       Docs: man:zram-generator(8)
             man:zram-generator.conf(5)
    Process: 9827 ExecStart=/lib/systemd/system-generators/zram-generator --setup-device zram1 (code=exited, status=0/SUCCESS)
    Process: 9832 ExecStop=/lib/systemd/system-generators/zram-generator --reset-device zram1 (code=exited, status=0/SUCCESS)
   Main PID: 9827 (code=exited, status=0/SUCCESS)

Jul 26 22:25:53 hswong3i-XPS-13-9300 zram-generator[9830]: Superblock backups stored on blocks:
Jul 26 22:25:53 hswong3i-XPS-13-9300 zram-generator[9830]:         32768, 98304, 163840, 229376, 294912, 819200, 884736, 1605632
Jul 26 22:25:53 hswong3i-XPS-13-9300 zram-generator[9830]: [49B blob data]
Jul 26 22:25:53 hswong3i-XPS-13-9300 zram-generator[9830]: [46B blob data]
Jul 26 22:25:53 hswong3i-XPS-13-9300 zram-generator[9830]: Creating journal (16384 blocks): done
Jul 26 22:25:53 hswong3i-XPS-13-9300 zram-generator[9830]: [83B blob data]
Jul 26 22:25:53 hswong3i-XPS-13-9300 systemd[1]: Finished Create swap on /dev/zram1.
Jul 26 22:25:53 hswong3i-XPS-13-9300 systemd[1]: Stopping Create swap on /dev/zram1...
Jul 26 22:25:53 hswong3i-XPS-13-9300 systemd[1]: systemd-zram-setup@zram1.service: Succeeded.
Jul 26 22:25:53 hswong3i-XPS-13-9300 systemd[1]: Stopped Create swap on /dev/zram1.
```
```
# journalctl -xef | egrep -e '^\w.*(systemd|zram)'
Jul 26 22:25:52 hswong3i-XPS-13-9300 kernel: zram1: detected capacity change from 0 to 16777216
Jul 26 22:25:52 hswong3i-XPS-13-9300 zram-generator[9830]: mke2fs 1.45.7 (28-Jan-2021)
Jul 26 22:25:53 hswong3i-XPS-13-9300 zram-generator[9830]: [120B blob data]
Jul 26 22:25:53 hswong3i-XPS-13-9300 zram-generator[9830]: Creating filesystem with 2097152 4k blocks and 524288 inodes
Jul 26 22:25:53 hswong3i-XPS-13-9300 zram-generator[9830]: Filesystem UUID: c8cf6c3f-6dd1-48e6-86a2-9f46ae887980
Jul 26 22:25:53 hswong3i-XPS-13-9300 zram-generator[9830]: Superblock backups stored on blocks:
Jul 26 22:25:53 hswong3i-XPS-13-9300 zram-generator[9830]:         32768, 98304, 163840, 229376, 294912, 819200, 884736, 1605632
Jul 26 22:25:53 hswong3i-XPS-13-9300 zram-generator[9830]: [49B blob data]
Jul 26 22:25:53 hswong3i-XPS-13-9300 zram-generator[9830]: [46B blob data]
Jul 26 22:25:53 hswong3i-XPS-13-9300 systemd[1]: update-notifier-download.service: Succeeded.
Jul 26 22:25:53 hswong3i-XPS-13-9300 systemd[1]: Finished Download data for packages that failed at package install time.
Jul 26 22:25:53 hswong3i-XPS-13-9300 zram-generator[9830]: Creating journal (16384 blocks): done
Jul 26 22:25:53 hswong3i-XPS-13-9300 zram-generator[9830]: [83B blob data]
Jul 26 22:25:53 hswong3i-XPS-13-9300 systemd[1]: Finished Create swap on /dev/zram1.
Jul 26 22:25:53 hswong3i-XPS-13-9300 systemd[1]: var-lib-libvirt-images.mount: Directory /var/lib/libvirt/images to mount over is not empty, mounting anyway.
Jul 26 22:25:53 hswong3i-XPS-13-9300 systemd[1]: Mounting Compressed Storage on /dev/zram1...
Jul 26 22:25:53 hswong3i-XPS-13-9300 systemd[1]: var-lib-libvirt-images.mount: Mount process exited, code=exited, status=1/FAILURE
Jul 26 22:25:53 hswong3i-XPS-13-9300 systemd[1]: var-lib-libvirt-images.mount: Failed with result 'exit-code'.
Jul 26 22:25:53 hswong3i-XPS-13-9300 systemd[1]: Failed to mount Compressed Storage on /dev/zram1.
Jul 26 22:25:53 hswong3i-XPS-13-9300 systemd[1]: Stopping Create swap on /dev/zram1...
Jul 26 22:25:53 hswong3i-XPS-13-9300 kernel: zram1: detected capacity change from 16777216 to 0
Jul 26 22:25:53 hswong3i-XPS-13-9300 systemd[1]: systemd-zram-setup@zram1.service: Succeeded.
Jul 26 22:25:53 hswong3i-XPS-13-9300 systemd[1]: Stopped Create swap on /dev/zram1.
```

Any idea why incorrectly detect `fs-type` as `swap`?

src/generator.rs
```diff
@@ -285,9 +285,14 @@ After=systemd-zram-setup@{zram_device}.service
 [Mount]
 What=/dev/{zram_device}
 Where={mount_point:?}
+Options={mount_options:?}
 ",
             zram_device = device.name,
             mount_point = device.mount_point.as_ref().unwrap(),
+            mount_options = match &device.mount_options {
```
nabijaczleweli, line 292: `unwrap_or()`, or, better yet, omit `Options=` if `mount_options` is `None`

nabijaczleweli (changes requested):

src/generator.rs
hswong3i: > omit Options= if mount_options is None

May I have some sample code snippet?

hswong3i:
@nabijaczleweli any more hits could be provided for https://github.com/systemd/zram-generator/pull/90#issuecomment-886755821?

keszybz:
Please drop the first commit from this pull request.

hswong3i:
> Please drop the first commit from this pull request.

Then the CI will fail due to missing kernel module, are you sure?

keszybz:
> > Please drop the first commit from this pull request.
> 
> Then the CI will fail due to missing kernel module, are you sure?

You are pulling in unrelated changes into the pull request. If the requirement for a new module is added *in this pull request*, then the commit to install the module should also be in this pull request. If not, then not. And the other changes to Makefiles make this pull request unmergeable. Please clean it up to keep only the relevant parts.

hswong3i:
> You are pulling in unrelated changes into the pull request. If the requirement for a new module is added _in this pull request_, then the commit to install the module should also be in this pull request. If not, then not. And the other changes to Makefiles make this pull request unmergeable. Please clean it up to keep only the relevant parts.

Could we make #92 and #93 get merged? I will follow up this PR afterward.

keszybz:
> Could we make #92 and #93 get merged? I will follow up this PR afterward.

Done.

hswong3i:
> > Could we make #92 and #93 get merged? I will follow up this PR afterward.
> 
> Done.

Done

src/generator.rs
```diff
@@ -283,11 +283,15 @@ Requires=systemd-zram-setup@{zram_device}.service
 After=systemd-zram-setup@{zram_device}.service
 
 [Mount]
+Where={mount_point}
 What=/dev/{zram_device}
-Where={mount_point:?}
```
keszybz, line 287: Please don't reverse the order here: we generally put What= before Where= in mount units, as this is the order on the `mount` commandline.

This is also doing an unrelated change to the formatting of Where. It's the same change as in #94. But please doing mixing unrelated changes into your pull requests.

src/generator.rs
```diff
@@ -283,11 +283,15 @@ Requires=systemd-zram-setup@{zram_device}.service
...
+Type={fs_type}
+Options={mount_options}
 ",
+            mount_point = device.mount_point.as_ref().unwrap().to_str().unwrap(),
             zram_device = device.name,
-            mount_point = device.mount_point.as_ref().unwrap(),
+            fs_type = device.fs_type.as_ref().unwrap(),
+            mount_options = device.mount_options.as_ref().unwrap_or(&String::new()),
```
keszybz, line 294: Please don't write empty "Options=", but instead skip the whole line if `device.mount_options` is None.

keszybz (changes requested):
This mostly looks reasonable, but:
1. please rebase on top of current `main`
2. please explain in the commit message what the change is. (Right now there's a discussion of why the change is useful, which is good, but cannot be the only content of the commit message.)
3. the new option needs to be described in the docs (at least in `man/zram-generator.conf.md`).
4. why are the test cases renamed? Either drop that change, or if there's some good reason for it, make it a separate commit and justify.

src/generator.rs
hswong3i: Errr..... I checked if create a CephFS mount in `/etc/fstab` with `_netdev`, the corresponding file ordered as my changes...

Also another example from my /boot/efi:
```
$ cat /run/systemd/generator/boot-efi.mount 
# Automatically generated by systemd-fstab-generator

[Unit]
Documentation=man:fstab(5) man:systemd-fstab-generator(8)
SourcePath=/etc/fstab
Requires=systemd-fsck@dev-disk-by\x2duuid-667C\x2d52E1.service
After=systemd-fsck@dev-disk-by\x2duuid-667C\x2d52E1.service
After=blockdev@dev-disk-by\x2duuid-667C\x2d52E1.target

[Mount]
Where=/boot/efi
What=/dev/disk/by-uuid/667C-52E1
Type=vfat
Options=umask=0077
```

src/generator.rs
hswong3i: I make it default with `defaults`

hswong3i:
> This mostly looks reasonable, but:
> 
>     1. please rebase on top of current `main`

Done
 
>     2. please explain in the commit message what the change is. (Right now there's a discussion of why the change is useful, which is good, but cannot be the only content of the commit message.)

Doesn't the commit title `mount-options: Add Mount Options Support` also enough?
 
>     3. the new option needs to be described in the docs (at least in `man/zram-generator.conf.md`).

Done

>     4. why are the test cases renamed? Either drop that change, or if there's some good reason for it, make it a separate commit and justify.

After reorder, the options provided is now increased from 1 -> 2 -> 3, which simpler for tracing:
```
$ cat tests/07-plain-device/etc/systemd/zram-generator.conf 
[zram11]
fs-type = ext4

$ cat tests/08-mount-point/etc/systemd/zram-generator.conf 
[zram11]
mount-point = /var/compressed
fs-type = ext4

$ cat tests/09-mount-options/etc/systemd/zram-generator.conf 
[zram11]
mount-point = /var/compressed
fs-type = ext4
mount-options = discard
```

src/generator.rs
hswong3i: > Please don't reverse the order here: we generally put What= before Where= in mount units, as this is the order on the mount commandline.

Ok agree with you (even systemd-fstab-generator is not working in this way), What -> Where -> Type -> Options also in the same order as `/etc/fstab` which looks much comfortable for me, too.

keszybz:
> 4. why are the test cases renamed? Either drop that change, or if there's some good reason for it, **make it a separate commit** and justify.

OK. But it needs to be a separate commit.

> Doesn't the commit title mount-options: Add Mount Options Support already enough?

The title doesn't explain what the option name is, and what it does. A person looking at the commit list should be able to figure out what a commit does from the description. A much better title would be "Add Options= setting to set mount options" and then have some more details in the commit body, for example a very simple example of the option and the effect on the generated unit.

src/generator.rs
keszybz: https://github.com/systemd/systemd/pull/20411

keszybz:
Please undo all the unrelated changes from commit. Let's try to do one thing first.

hswong3i:
> Please undo all the unrelated changes from commit, like the file system type and default options. Let's try to do one thing first correctly.

Done, also split PR to #95 and #96

man/zram-generator.conf.md
```diff
@@ -104,6 +104,12 @@ Devices with the final size of *0* will be discarded.
 
   Also see systemd-makefs(8).
 
+* `mount-options` =
+
+  Specifies how the device shall be mount with options. The default is *discard* for enable TRIM support for ext4.
```
keszybz, line 109: Please drop the part about chaning the default options for now.

keszybz (changes requested):

man/zram-generator.conf.md
hswong3i: Done

hswong3i:
> > 1. why are the test cases renamed? Either drop that change, or if there's some good reason for it, **make it a separate commit** and justify.
> 
> OK. But it needs to be a separate commit.

See #96 

> > Doesn't the commit title mount-options: Add Mount Options Support already enough?
> 
> The title doesn't explain what the option name is, and what it does. A person looking at the commit list should be able to figure out what a commit does from the description. A much better title would be "Add Options= setting to set mount options" and then have some more details in the commit body, for example a very simple example of the option and the effect on the generated unit.

Done

PR #92, opened 2021-07-31
.github/workflows/ci.yml: Install linux-modules-extra

hswong3i:
GitHub Actions is now broken due to missing zram kernel module, therefore modprobe failed.

This PR install `linux-modules-extra-$(uname -r)` so auto detect running kernel version and install corresponding required kernel modules correctly.


keszybz:
Indeed, this installs `linux-modules-extra-5.8.0-1039-azure`, which has `zram.ko`.

PR #95, opened 2021-08-10
fs-type: Switch Default from ext2 to ext4

hswong3i:
When using zram for file system, we need to enable `discard` mountoptions for issue discard/TRIM commands to the underlying zram blockdevice when blocks are freed.
    
This PR switch our default `fs-type = ext2` to `fs-type = ext4` forabove `discard` mount options support.
    

keszybz:
It seems ext2 supports discard. I mounted an ext2 fs and then did a manual `sudo mount -o discard,remount /tmp/mountpoint` 
and I see that with the discard option, zram memory usage goes down as expected. Without `discard`, it stays up even after files have been removed. So the premise that we need to change to ext4 to support discard is wrong.

Thus, ext2 should stay as the default. Instead, the default should be to use discard.

We can add an Options= setting, it could be useful for other things.

PR #96, opened 2021-08-10
test_cases.rs: Reorder Test Cases

hswong3i:
test_cases.rs: Reorder Test Cases

After test cases reordered, the testing parameters for file systemd related are now increasing in step by step for simpler management, e.g.

    $ cat tests/07-plain-device/etc/systemd/zram-generator.conf
    [zram11]
    fs-type = ext4

    $ cat tests/08-mount-point/etc/systemd/zram-generator.conf
    [zram11]
    mount-point = /var/compressed
    fs-type = ext4

    $ cat tests/09-mount-options/etc/systemd/zram-generator.conf
    [zram11]
    mount-point = /var/compressed
    fs-type = ext4
    mount-options = discard


keszybz:
There are 3 extra commits in this PR that don't belong here.

PR #97, opened 2021-08-13
Describe how zram-generator is invoked after install in readme

tomyo:
Took me some time to figure out what to do after install, reading man pages I understood is executed automatically on boot.
I think it would be nice to be explicit on that on the readme.


vilgotf:
Maybe instead of trying to describe it we could link to this https://www.freedesktop.org/software/systemd/man/systemd.generator.html?

keszybz:
I merged this manually with some edits as 771df17d97.

PR #105, opened 2021-10-14
man: describe how to apply config changes at runtime

### LINKED ISSUE #78: Apply config changes?
[In Jan 2021 I briefly mentioned this concern](https://github.com/systemd/zram-generator/pull/60).

> Ideally the proper way to handle this refresh/re-initialization/update is to **add documentation for how users should apply config updates.**

Is this supported? It doesn't appear to be documented.

How do I adjust a config and apply the changes? (_such as changing number of devices or their config options and values_)

Do I need to restart the system / session?

keszybz:
Fixes #78.

@polarathene PTAL.

### DIFF
--- a/man/zram-generator.md
+++ b/man/zram-generator.md
@@ -60,13 +60,13 @@ Otherwise, systemd will not know how to stop the unit properly.
 
 ## REPORTING BUGS
 
-&lt;<https://github.com/systemd/zram-generator/issues>&gt;
+<https://github.com/systemd/zram-generator/issues>
 
 ## SEE ALSO
 
 zram-generator.conf(5), systemd.generator(7), systemd.swap(5)
 
-&lt;<https://github.com/systemd/zram-generator>&gt;
+<https://github.com/systemd/zram-generator>
 
-Linux documentation of zram: &lt;<https://kernel.org/doc/html/latest/admin-guide/blockdev/zram.html>&gt;<br />
-     and the zram sysfs ABI: &lt;<https://kernel.org/doc/Documentation/ABI/testing/sysfs-block-zram>&gt;
+Linux documentation of zram: <https://kernel.org/doc/html/latest/admin-guide/blockdev/zram.html><br />
+     and the zram sysfs ABI: <https://kernel.org/doc/Documentation/ABI/testing/sysfs-block-zram>


keszybz:
I applied the suggestions above, except as commented. And I added a second commit to remove some formatting.

keszybz:
This pull request is essentially a single paragraph. And when reviewing changes to text you generally need to look at the whole paragraph, to control for style consistency, lack of repeats, clarity, etc. So splitting the change into multiple commits doesn't actually help with review. (It would only be useful for trivial things like you point a typo and I fix a typo. But we don't need a re-review for those changes.)

Also, when reviewing pull requests, I want to be able to review the clarity of the split into commits, and commit messages themselves. With history which consists of real commits and fixup commits you can't really do this.

But more generally, in my experience, pull requests with multiple "as you go" commits lead to those fixup commits being merged into history. Once the pull request is done one would have to go back, and split the fixup commits into pieces which apply to the "real" commits. But rebasing is hard, and I'm likely to introduce some errors then.

So overall, if at the end you want to have clean history, I think it's better to maintain such state at all times. It may make the review a bit harder in some cases, but in other cases it is either neutral or actually makes review easier.

keszybz:
I'll merge this as is. I think it's good enough, and the proposed changes are just a matter of personal preferences.

polarathene:
Review finished (despite missing the merge):

- `filesystem` appears to be prevalent in systemd still: https://github.com/systemd/zram-generator/pull/105#discussion_r732295179
- Received no feedback for change requesting adding caution on restarting zram swap device if lacking sufficient memory: https://github.com/systemd/zram-generator/pull/105#discussion_r732291653
- Suggestion was dismissed for reason that made little sense when suggested revision offered many other benefits over original content: https://github.com/systemd/zram-generator/pull/105#discussion_r732276589
- Feedback was given that didn't make much sense, no response received when questioned: https://github.com/systemd/zram-generator/pull/105#discussion_r732293362
- Advice to reduce friction on reviewers by appending commits during review instead of rewriting history (comment prior to this one): https://github.com/systemd/zram-generator/pull/105#issuecomment-947126212

---

Please do not ping me for review in future if you're going to dismiss the bulk of it and are unwilling to be accommodating when making changes.

From what I could see, very little was actually changed in the changes that I had to manually lookup commits and diff:

```diff
--- https://github.com/systemd/zram-generator/commit/0a3c88f236817137d85432397f7d0ff36408d28c
+++ https://github.com/systemd/zram-generator/commit/40aa017abd69b72a066b8ca87e6f2eb5c1c23494
@@ -5,10 +5,11 @@
 
 Nevertheless, sometimes it may be useful to add new devices or apply config changes at runtime.
 Applying new configuration means restarting the units, and that in turn means recreating the zram devices.
-This means that filesystems are temporarily unmounted and their contents lost, and pages are moved out of swap devices into other memory.
-If this is acceptable, `systemctl restart systemd-zram-setup@zram<N>` or `systemctl restart systemd-zram-setup@*`
+This means that *file systems are temporarily unmounted and their contents lost*, and *pages are moved out of the compressed swap device* into other memory.
+If this is acceptable, `systemctl restart systemd-zram-setup@zramN` or `systemctl restart systemd-zram-setup@*`
 may be used to recreate a specific device or all configured devices.
+(If the device didn't exist, `restart` will create it.)
 If the way the device is used (e.g. the mount point or file system type) is changed,
-`systemctl daemon-reload` needs to be called first to recreate `systemd` units.
+`systemctl daemon-reload` needs to be called first to recreate systemd units.
 If a device or mount point is removed from configuration, the unit should be stopped before calling `daemon-reload`.
 Otherwise, systemd will not know how to stop the unit properly.
```

PR #109, opened 2021-10-19
Use actual comment for SPDX comment in ronn source, match link Aqs

nabijaczleweli:
This removes it from the GitHub rich view


keszybz:
Ah, indeed. I googled for 'markdown comment', and got that. But your version seems to work better and is more obvious. (Ghh, why is markdown so unspecified.)

PR #121, opened 2021-10-27
Replace zram-fraction and max-zram-size with expression zram-size

### LINKED ISSUE #51: RFE: support boosting zram-fraction on low memory machines

nabijaczleweli:
Closes #51 because you can either do PWL or logarithmic.
Individual commits atomic, each one passes CI. Needs manual rebase on top of #120.

`fasteval` isn't the most active of implementations, but it's well-thought-through, small, and, well, not Explicitly Abandoned.

Sizing, current HEAD  (bd5c140ee65125e821214bc09e059a3cc4872f1c) against tip of this loses just 23k:
```
$ l pre post
-rwxr-xr-x 1 nabijaczleweli users 855K Nov  2 13:10 pre
-rwxr-xr-x 1 nabijaczleweli users 903K Nov  2 13:10 post

$ size pre post
   text    data     bss     dec     hex filename
 826535   42441     776  869752   d4578 pre
 868863   45537     776  915176   df6e8 post

$ du -h pre post
421K    pre
445K    post
```

~~This *is* breaking! I didn't want to argue about tagging off 1.x despite no broken interfaces but now I realise I definitely should've!
There, by design, is an easy cleave point: before (or after, either works) the fourth commit. I'm posting this as a single PR for initial review, but I propose:~~
  * ~~pulling in the first four commits~~
  * ~~bumping to 1.1.0~~
  * ~~pulling in the rest~~
  * ~~bumping to 2.0.0~~

~~That way people can spec 1.x to port off the deprecated interfaces, and upgrade to 2.x otherwise.
I can either inject the 1.1 bump into here or split it off into two PRs, if desired.~~


nabijaczleweli:
Re-phrased these and added some more tests. The short-hand functional delta off the last version is I moved the obsolete options to their own section, added a test that we actually parse the obsolete values, and squished them in the example.

keszybz:
Looks great.

keszybz (approved):

Issue #122, opened 2021-10-29
RFE: add support for backing device


fignew:
From https://www.kernel.org/doc/html/latest/admin-guide/blockdev/zram.html#optional-feature

Zram can write idle/incompressible page to backing storage rather than keeping it in memory.

This is a useful feature that extends the functionality of zram and should be configurable via zram-generator.

Thanks

keszybz:
I guess this could be done… We should extend the unit to depend on **two** devices: the zram device and the backing device. The backing device would be specified with any method supported by systemd. When the backing device appears, we would write the device path in a format that the kernel understands to /sys.

PR #124, opened 2021-11-02
Add writeback-dev=, corresponding to /s/b/z0/backing_dev

nabijaczleweli:
CC: @fignew, please test if this works in your setup

Also, fun fact, the documentation says, i.a. "If admin wants to use incompressible page writeback, they could do via:" – this is a lie! Those are all triggers, not toggles. This is all we can configure.


nabijaczleweli:
rebased

src/setup.rs
```diff
@@ -42,6 +43,16 @@ pub fn run_device_setup(device: Option<Device>, device_name: &str) -> Result<()>
         }
     }
 
+    if let Some(ref wb_dev) = device.writeback_dev {
+        let writeback_path = device_sysfs_path.join("backing_dev");
+        fs::write(&writeback_path, wb_dev.as_os_str().as_bytes()).with_context(|| {
+            format!(
+                "Failed to configure write-back device into {}",
+                writeback_path.display()
```
keszybz, line 51: I'm worried about this part… As I mentioned in the discussion on #122, the user can specify a path like `/dev/disk/by-uuid/abcd-1234` and I'm not sure if the kernel will know that this is the same as `/dev/sda5`.

Unfortunately this feature requires `CONFIG_ZRAM_WRITEBACK`, and Fedora kernels are compiled without that. So I'll need to first recompile the kernel to test this :(

nabijaczleweli: Yes, and I tried that (well, I used a /dev/zvol/ path, but these also symlinks to ../zdN), and it Just Works (the file's opened with `backing_dev = filp_open(file_name, O_RDWR|O_LARGEFILE, 0)`, which literally just decays to what userspace open(2) does).
On read-back, `backing_dev` returns the real `/dev/zdN` path, but that does not matter at all.

Cf. transcript:
```
# zfs create -sV 40G tarta-zoot/test-wb
# modprobe zram
# tee /sys/block/zram0/backing_dev
/dev/zram/tarta-zoot/test-wb
/dev/zram/tarta-zoot/test-wb
tee: /sys/block/zram0/backing_dev: No such file or directory
^C
# tee /sys/block/zram0/backing_dev
/dev/zvol/tarta-zoot/test-wb
/dev/zvol/tarta-zoot/test-wb
$ cat /sys/block/zram0/backing_dev
/dev/zd176
```

keszybz:
Two issues I noticed in testing:
- I think this also needs an `After=<backing-device>`. I didn't test it, but w/o After, we could try to write the name too early, before the device node exists and the kernel can resolve the path to a device. On shutdown, systemd could also try to destroy the backing device before the zram device.
- There is a disagreement about escaping: I tried `/dev/disk/by-path/virtio-pci-0000:08:00.0`, and we escape this as "dev-disk-by\\x2dpath-virtio\\x2dpci\\x2d0000\\x3a08\\x3a00.0.device", but systemd has "dev-disk-by\\x2dpath-virtio\\x2dpci\\x2d0000:08:00.0.device". I didn't yet figure out who misimplemented the spec, but I expect that it doesn't matter: the systemd implementation is probably the de-facto "standard", and even if the docs disagree, they might be adjusted to fit it.

keszybz:
Oh, no, we get this wrong. OK, so this is an easy fix, we just need to adjust the algorithm that was added recently.

src/generator.rs
```diff
@@ -228,14 +249,14 @@ Options={options}
 /// Path escaping as described in systemd.unit(5)
 ///
 /// `/./` components stripped away when parsing `mount-point =`
-fn mount_unit_name(path: &Path) -> String {
+fn mount_unit_name(path: &Path, suffix: &str) -> String {
```
keszybz, line 247: I think this should be renamed to `unit_name_from_path()` now.

As discussed above, this also needs a following fixup:
```diff
-                b'0'..=b'9' | b'a'..=b'z' | b'A'..=b'Z' | b'_' | b'.' => obuf.push(b),
+                b'0'..=b'9' | b'a'..=b'z' | b'A'..=b'Z' | b'_' | b'.' | b':' => obuf.push(b),
```
and please test that `"/dev/disk/by-path/virtio-pci-0000:08:00.0"` expands to `"dev-disk-by\\x2dpath-virtio\\x2dpci\\x2d0000:08:00.0.device"`.

keszybz: (separate commit please, for clarity)

man/zram-generator.conf.md
```diff
@@ -71,6 +71,13 @@ Devices with the final size of *0* will be discarded.
 
   If unset, none will be configured and the kernel's default will be used.
 
+* `writeback-dev`=
+
+  Write incompressible pages, for which no gain was achieved, to the specified block device under memory pressure.
```
keszybz, line 76: Maybe "writeback-device" ? We don't use abbreviations in other option names?

I think some more explanation what the valid values are is needed?
```
"Takes a path to a block device, e.g. `/dev/disk/by-partuuid/22ecf232-c8a3-4627-b726-72f84d43de82` or `/dev/mapper/some-device/`."
```
?

nabijaczleweli:
Oh, lol, I was pretty sure I got this right, and I did! For systemd 247.3-6 (bullseye):
screenshot of man systemd.unit against systemd.unit.xml from the systemd repo
Apparently this got fixed in https://github.com/systemd/systemd/commit/aa6dc3ec337b04308a5dfe3b962fa88088b2c82e, which github says landed in v249

keszybz:
Oh, indeed. So it's as I wrote… systemd was in the wrong, but docs got adjusted ;) For us here that's good, because the code will with new systemd and old both.

src/generator.rs
nabijaczleweli: Fixed

man/zram-generator.conf.md
nabijaczleweli: "writeback-device" definitely, I was expecting many more options to use it as a prefix from the kernel documentation, but it turned out that was a lie, and didn't elongate it back.

Added the blurb, too.

nabijaczleweli:
As for the `After=`, you're probably right, since we could try to use it before the device is Active (what this strictly entails is beyond me for the time being), but fixed that, too.

nabijaczleweli:
Rebased

nabijaczleweli:
Friday bump :0 I think we've ironed out all the kinks out of this now?

src/generator.rs
```diff
@@ -150,16 +151,24 @@ fn write_contents(output_directory: &Path, filename: &str, contents: &str) -> Re
...
+                    .as_deref()
+                    .unwrap_or_else(|_| Path::new(concat!(crate_name!(), " ", crate_version!())))
+                    .as_os_str()
+                    .as_bytes(),
+            )?;
+            f.write_all(b"\n\n")?;
+            f.write_all(contents.as_bytes())?;
+            f.flush()
+        })
+        .with_context(|| format!("Failed to write {}", path.display()))
```
keszybz, line 172: I don't find this to be an improvement… The code is longer and **much** harder to read. If this was in some hot path, then sure, it'd make sense. But we generally create a few files, and even if we "overallocate" here, this doesn't matter at all.

src/generator.rs
```diff
@@ -157,13 +164,7 @@ fn write_contents(output_directory: &Path, filename: &str, contents: &str) -> Re
...
         .and_then(|mut f| {
             f.write_all(b"# Automatically generated by ")?;
-            f.write_all(
-                env::current_exe()
-                    .as_deref()
-                    .unwrap_or_else(|_| Path::new(concat!(crate_name!(), " ", crate_version!())))
-                    .as_os_str()
-                    .as_bytes(),
-            )?;
+            f.write_all(*CURRENT_EXE)?;
```
keszybz, line 167: Same story here. The code is not longer, but it does bring in a new dep, and is harder to read. So we're making future maintenance harder, and we don't really gain much. In a debug build this adds 130kb of text.

keszybz:
Seems to work great. I'll go ahead and merge the first three commits so this can get more testing. We can discuss the other two commits, but I'm not in favour, as described above.

src/config.rs
```diff
@@ -45,6 +46,7 @@ impl Device {
             host_memory_limit_mb: None,
             zram_size: None,
             compression_algorithm: None,
+            writeback_dev: None,
```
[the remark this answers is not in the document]
nabijaczleweli: meh. i mean, sure, but bit late for that now.

nabijaczleweli:
yeah, these would be gains if the features I wanted were actually stable, the dependency on `once-cell` is meh since it's being stabilised verbatim, but it's taking time, and `writev()` would be beneficial since we already have the buffers, but we end up allocating into BufWriter anyway, so maybe Next Year™

PR #127, opened 2021-11-07
Fixups for the expression pull request

keszybz:


src/main.rs
```diff
@@ -45,7 +45,7 @@ fn get_opts() -> Opts {
         .get_matches();
 
     let val = opts
-        .value_of("directory/device")
+        .value_of("directory|device")
```
nabijaczleweli, line 48: oops!

src/config.rs
```diff
@@ -359,10 +354,8 @@ fn parse_line(dev: &mut Device, key: &str, value: &str) -> Result<()> {
         }
 
         "zram-fraction" => {
-            warn!(
-                "{}: deprecated {} configuration key set. Use {} = instead!",
-                dev.name, key, "zram-size"
-            );
+            /* zram-fraction is for backwards compat. zram-size = is preferred. */
```
nabijaczleweli, line 357: Why? If you don't actively push people away from this they won't fix their configs. Systemd itself has whined on each reload about tiny non-problems for years, and this is not that.

keszybz: The point is that I don't see much value in forcing people to "fix" their configs. There is simply nothing *wrong* with their configs. Systemd has a bit of a varied record here: in the past there was certainly quite a lot of push in such a direction, but nowadays we don't really do that. And in particular for config options and command-line option names, endless and silent backwards compat is the norm. Systemd will push people to update, but nowadays mostly when there's a particular reason. I listed some cases in the commit message. I think none of that applies here, and in particular I don't think we should "break" (in the sense of requiring adjustments) the default configs that were widely advertised when zram-generator was made the default in Fedora.

nabijaczleweli: Meh, sure, I don't agree necessarily but I'll just run the -R of this commit so v0v.

But as a separate point: does Fedora not have, like, a ucf alternative?

keszybz: No, Fedora uses a different approach. Generally packages are supposed to preserve compatibility, and if not possible, do changes using automatic scriptlets. In general we try to make /etc as empty as possible and avoid endless configuration of everything.

PR #128, opened 2021-11-08
More make shenanigans

keszybz:


Makefile
```diff
@@ -6,6 +6,7 @@ CARGOFLAGS =
 RONN = ronn
 PKG_CONFIG = pkg-config
 PREFIX = /usr
+BUILDTYPE = release
```
nabijaczleweli, line 9:
```
BUILDTYPE ?= release
CARGOFLAGS += --$(BUILDTYPE)
```
will make the body cleaner and fix clippy being build in debug by default

keszybz:
PTAL again.

nabijaczleweli (approved):

nabijaczleweli (changes requested):
Actually, sorry, on reexamining the CI config, I was wrong to suggest `CARGOFLAGS += --$(BUILDTYPE)` and approved this too soon when I saw the short diff.
It should be `CARGOFLAGS := --$(BUILDTYPE) $(CARGOFLAGS)` — it's valid, correct, and expected to pass arguments to the programs Cargo runs. Look no further than
```
      - name: Validate clippy
        run: SYSTEMD_UTIL_DIR=/ cargo clippy -- -D warnings
```

It's impossible to express it *at all* if we append, but if we pre-pend, it's trivial (and actually useful) as
```
      - name: Validate clippy
        run: make clippy CARGOFLAGS='-- -D warnings'
```
which will, indeed, run `cargo clippy --release -- -D warnings`.

And `make check CARGOFLAGS='--verbose -- --nocapture` correctly runs `cargo test --release --verbose -- --nocapture`.

Makefile
vilgotf: I'd prefer defaulting to `debug`, I think rust devs would expect that

keszybz:
> `CARGOFLAGS := --$(BUILDTYPE) $(CARGOFLAGS)`

I had it this, but changed to the shorter form for simplicity…

nabijaczleweli:
ya, that's good, but I'd like the first commit to also use the `make clippy CARGOFLAGS='-- -D warnings' stanza in the GHA config (maybe I shoulda made that a stronger point in my last comment)

nabijaczleweli:
Oh, also, there does seem to be a useful difference between `DEFAULT = dfl`:
```
nabijaczleweli@tarta:~/uwu$ cat Makefile
DEFAULT = dfl
_:
        @echo $(DEFAULT)
nabijaczleweli@tarta:~/uwu$ make
dfl
nabijaczleweli@tarta:~/uwu$ make DEFAULT=a
a
nabijaczleweli@tarta:~/uwu$ DEFAULT=a make
dfl
```
vs:
```
nabijaczleweli@tarta:~/uwu$ cat Makefile
DEFAULT ?= dfl
_:
        @echo $(DEFAULT)
nabijaczleweli@tarta:~/uwu$ make
dfl
nabijaczleweli@tarta:~/uwu$ make DEFAULT=a
a
nabijaczleweli@tarta:~/uwu$ DEFAULT=a make
a
```

Which means that the variable just gets quietly eaten if you use `=`!
The manual [backs this up](https://www.gnu.org/software/make/manual/make.html#Setting) (6th paragraph down):
> If you’d like a variable to be set to a value only if it’s not already set, then you can use the shorthand operator ‘?=’ instead of ‘=’.

In retrospect, this does explain why the makefile didn't seem to work right when I first tried it.


As an additional Fun Time, we are [forbidden from modifying arguments](https://www.gnu.org/software/make/manual/make.html#Override-Directive), unless we use `override`, which seems so obscure vim doesn't even highlight it, [the assignment gets ignored](https://github.com/systemd/zram-generator/runs/4155060954?check_suite_focus=true#step:4:14).


In the process of writing and testing this comment (and GNU make's own documentation), I've made a branch and it seems to have reached functional equilibrium, so instead of blabbing more I'll just open a PR.

PR #129, opened 2021-11-08
Anti --ignore-matching-lines aktion

nabijaczleweli:
~~Currently on top of #124, and see below for top commit keep-in vs removal, hence draft~~ Rebased, removed

This removes all `--ignore-matching-lines`: the `Exec*=` ones were unused, the second we, on init (i.e. before the runner clones threads, which prevents userns creation), make a user- and mountns, become root in it, and mount over tmpfs to fake /proc/self/exe.

I guess for super-duper cleanliness we could setugid back to originals, but I don't think it matters at all, so.

The dev-dep graph grows to:
```
[dev-dependencies]
├── ctor v0.1.21 (proc-macro)
│   ├── quote v1.0.10
│   │   └── proc-macro2 v1.0.32
│   │       └── unicode-xid v0.2.2
│   └── syn v1.0.81
│       ├── proc-macro2 v1.0.32 (*)
│       ├── quote v1.0.10 (*)
│       └── unicode-xid v0.2.2
└── nix v0.23.0
    ├── bitflags v1.3.2
    ├── cfg-if v1.0.0
    ├── libc v0.2.105
    └── memoffset v0.6.4
        [build-dependencies]
        └── autocfg v1.0.1
```

On top, there's a port to pure libc, which reduces the graph to (libc is already part of our graph a few times):
```
[dev-dependencies]
├── ctor v0.1.21 (proc-macro)
│   ├── quote v1.0.10
│   │   └── proc-macro2 v1.0.32
│   │       └── unicode-xid v0.2.2
│   └── syn v1.0.81
│       ├── proc-macro2 v1.0.32 (*)
│       ├── quote v1.0.10 (*)
│       └── unicode-xid v0.2.2
└── libc v0.2.105
```
if we care about dev-dependencies graph (which I don't think we do, but), but it kinda sucks, more so with the formatting, so I'd rather just go with nix


keszybz:
This is pretty nice. I think the extra deps are OK: those are both fairly widely used packages, both are available in Fedora. So I'd prefer to drop the last two commits (or at least the last one).

nabijaczleweli:
That's great news, dropped the top two.

Old HEAD (without `libc` and `ctor`) at https://github.com/nabijaczleweli/systemd-zram-generator/tree/ignore-matching-lines%2Bfangs, if it's needed again for some reason.

PR #130, opened 2021-11-09
More make shenanigans (redux)

nabijaczleweli:
Supersedes and includes #128 – with ?= at the front, rebased, and override := in the middle.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,4 +73,4 @@ jobs:
           profile: minimal
           components: clippy
       - name: Validate clippy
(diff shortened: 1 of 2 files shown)


keszybz:
I'm meh about the change to ?=: I think that `make FOO=…` *is* better than `FOO=… make`. But OTOH, my opinion is not strong enough to oppose this. Let's merge.

nabijaczleweli:
Whether it's better is secondary. That it *doesn't work* when you configure a global alternate cargo or any other program override (cf. exporting CC=clang-11 or w/e) is the main concern, really.

PR #138, opened 2021-12-01
Skip test if unshare fails

keszybz:
@nabijaczleweli This works, but isn't very pretty. Maybe you have a better idea.

### DIFF
--- a/tests/test_cases.rs
+++ b/tests/test_cases.rs
@@ -6,18 +6,26 @@ use anyhow::Result;
 use fs_extra::dir::{copy, CopyOptions};
 use std::fs;
 use std::io::{self, Write};
+use log::warn;
 use std::path::Path;
-use std::process::Command;
+use std::process::{Command, exit};
 use tempfile::TempDir;
 
 #[ctor::ctor]
 fn unshorn() {
-    use nix::{mount, sched, unistd};
+    use nix::{errno, mount, sched, unistd};
     use std::os::unix::fs::symlink;
 
     let (uid, gid) = (unistd::geteuid(), unistd::getegid());
     if !uid.is_root() {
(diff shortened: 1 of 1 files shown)


tests/test_cases.rs
```diff
@@ -6,18 +6,26 @@ use anyhow::Result;
...
+    use nix::{errno, mount, sched, unistd};
     use std::os::unix::fs::symlink;
 
     let (uid, gid) = (unistd::geteuid(), unistd::getegid());
     if !uid.is_root() {
-        sched::unshare(sched::CloneFlags::CLONE_NEWUSER).expect("unshare(NEWUSER)");
+        match sched::unshare(sched::CloneFlags::CLONE_NEWUSER) {
+            Ok(_) => {}
+            Err(errno::Errno::EPERM) => {
+                warn!("No permission to unshare namespace, skipping tests");
```
nabijaczleweli, line 24: `warn!()` doesn't work here (there is no log sink installed) — just use `eprintln!()`.
Also, the message is odd (which namespace?) – how about "unshare(NEWUSER) forbidden and not running as root: skipping tests."?

tests/test_cases.rs
```diff
@@ -6,18 +6,26 @@ use anyhow::Result;
...
+                exit(0);
+            }
+            Err(err) => panic!("Unexpected error: {}", err),
```
nabijaczleweli, line 27: Again, what caused this error? Why not just
```rust
e @ Err(_) => e.expect("unshare(NEWUSER)"),
```
or, indeed, drop the oddly-phrased `Ok(()) => (),` case at the top and replace this line with just
```rust
r => r.expect("unshare(NEWUSER)"),
```

nabijaczleweli (changes requested):
Meh, as it goes, this is a sensible approach. Where are you running into this?

PR #143, opened 2021-12-07
Skip test if unshare fails (redux)

nabijaczleweli:
Supersedes #138 – this is my suggested edit from the comment, and warning- + rustfmt-clean.

I'm still curious where you've run into this.


keszybz:
> I'm still curious where you've run into this.

In Fedora koji: builds are running in some systemd-nspawn container that is fairly tightly locked down. It doesn't happen if I do something similar locally.

PR #145, opened 2021-12-10
Issue warning if /s/b/z0/backing_dev doesn't exist and wb-dev= set

nabijaczleweli:
I run Debian, so untested!


keszybz:
Looks like this should DTRT. Thanks!

PR #160, opened 2022-08-30
Include link to puppet-zram_generator

traylenator:
Include a link to the recently published puppet-zram-generator

https://github.com/voxpupuli/puppet-zram_generator

Mostly just letting you know - will not be sad if not merged.


keszybz:
Nice, thanks!

PR #171, opened 2022-12-03
fixup: reported improvements by Clippy

pheiduck:
Added in: **1.65.0**
`uninlined_format_args`
**What it does?**
Detect when a variable is not inlined in a format string, and suggests to inline it.
**Why is this bad?**
Non-inlined code is slightly more difficult to read and understand, as it requires arguments to be matched against the format string. The inlined syntax, where allowed, is simpler.

**Examples:**

```
format!("{}", var);
format!("{v:?}", v = var);
format!("{0} {0}", var);
format!("{0:1$}", var, width);
format!("{:.*}", prec, var);
```
Use instead:
```
format!("{var}");
format!("{var:?}");
format!("{var} {var}");
format!("{var:width$}");
format!("{var:.prec$}");
```

If allow-mixed-uninlined-format-args is set to false in clippy.toml, the following code will also trigger the lint:
`format!("{} {}", var, 1+2);`
Use instead:
`format!("{var} {}", 1+2);`

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -139,7 +139,7 @@ impl fmt::Display for Device {
         if self.zram_fraction.is_some() || self.max_zram_size_mb.is_some() {
             f.write_str(" (")?;
             if let Some(zf) = self.zram_fraction {
-                write!(f, "zram-fraction={}", zf)?;
+                write!(f, "zram-fraction={zf}")?;
             }
             if self.max_zram_size_mb.is_some() {
                 f.write_str(" ")?;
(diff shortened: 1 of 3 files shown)


nabijaczleweli:
calling it an "error" is FUD, it's a stylistic choice, and it necessarily drastically cuts off MSRV (in this case: past bullseye cargo/rustc)

pheiduck:
> calling it an "error" is FUD, it's a stylistic choice, and it necessarily drastically cuts off MSRV (in this case: past bullseye cargo/rustc)

Agree, it introduce build regressions as well so, we will see how this endup in future releases of rust.

pheiduck:
I would leave this open to work on.

pheiduck:
@nabijaczleweli you can review it again, I will squash my commits later to 1.

pheiduck:
Fixed are no longer necessary I will close this PR.
https://github.com/systemd/zram-generator/commit/af6fffe43a24b335abef3fd0f9b0bffdee4ef4c0 fixes some of them

PR #175, opened 2023-02-03
Use liboverdrop 0.1.0

nabijaczleweli:
edition=2021 required for `[].into_iter()` actually moving

This reduces needless allocations greatly :)

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -3,10 +3,10 @@
 use anyhow::{anyhow, Context, Result};
 use fasteval::Evaler;
 use ini::Ini;
-use liboverdrop::FragmentScanner;
 use log::{info, warn};
 use std::borrow::Cow;
 use std::collections::{BTreeMap, HashMap};
+use std::ffi::OsString;
 use std::fmt;
 use std::fs;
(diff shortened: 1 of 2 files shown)


src/config.rs
```diff
@@ -248,35 +248,29 @@ fn read_devices(
...
-        String::from(root.join("etc").to_str().unwrap()),
-        String::from(root.join("run").to_str().unwrap()), // We look at /run to allow temporary overriding
-                                                          // of configuration. There is no expectation of
-                                                          // programatic creation of config there.
+fn locate_fragments(root: &Path) -> BTreeMap<OsString, PathBuf> {
+    let base_dirs = [
+        root.join("usr/lib"),
+        root.join("usr/local/lib"),
+        root.join("etc"),
+        root.join("run"), // We look at /run to allow temporary overriding
```
cgwalters, line 256: This could use https://github.com/coreos/liboverdrop-rs/pull/26 in the future too.

cgwalters (approved):
Looks sane to me!

keszybz:
Looks reasonable. I don't think we need to care about compat with such old rust, so let's just do this.

PR #176, opened 2023-02-11
Fixup: useless conversion to the same type: `std::str::Lines<'_>`

pheiduck:
consider removing `.into_iter()`


nabijaczleweli:
yep

PR #180, opened 2023-05-02
CI: optimize removal of older rust binaries

pheiduck:
one `rm -rf ...` line instead of 2 or more later.
Before:
```
rm -f /home/runner/.cargo/bin/*fmt
rm -f ~/.cargo/bin/*fmt ~/.cargo/bin/rust-analyzer
```
After:
```
rm -f ~/.cargo/bin/{*fmt,rust-analyzer}
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,7 +34,7 @@ jobs:
         run: sudo modprobe -v zram
       - name: Install Rust
         run: |
-          rm -f ~/.cargo/bin/*fmt ~/.cargo/bin/rust-analyzer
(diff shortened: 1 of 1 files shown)


nabijaczleweli:
The actual change is
```diff
-rm -f ~/.cargo/bin/*fmt ~/.cargo/bin/rust-analyzer
+rm -f ~/.cargo/bin/{*fmt,rust-analyzer}
```
which is (a) the same line and process count (=> what's the optimisation here?) and (b) adds a kshism to (c) generate the same words for (d) no reason at all?

PR #184, opened 2023-08-18
cache: linux-modules-extra

pheiduck:
This method will reduce failed downloads of this package


nabijaczleweli:
Looks like speeds up the job by 10s.

PR #190, opened 2024-01-27
ci.yml: use a more up-to-date action to cache apt pkgs

pheiduck:


pheiduck:
we should stay on the current one, this one is not working

PR #193, opened 2024-04-08
ci.yml: remove any pre-existing rust tools first

pheiduck:
rustup can not handle the pre-existing rust bins during update so they have to be removed.
We can skip this by removing all pre-existing rust bins and install them via rustup

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,7 @@ jobs:
           sudo modprobe -v zram
       - name: Install Rust
         run: |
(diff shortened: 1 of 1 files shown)


keszybz:
Can you please provide a short description of the issue this is solving in the commit message?

pheiduck:
> Can you please provide a short description of the issue this is solving in the commit message?

Done.

pheiduck:
I checked ci output again and it does not really solve anything due there no issue in this case. So closed

PR #194, opened 2024-04-08
ci.yml: modeprobe zstd for tests

pheiduck:


.github/workflows/ci.yml
```diff
@@ -44,6 +44,10 @@ jobs:
           rustup default ${{ matrix.rust }}
       - name: Build
         run: make program CARGOFLAGS="--verbose"
+      - name: Insert zstd module
```
keszybz: Can you move it up right after "Insert zram module"? I think it's nicer to keep those two together.

.github/workflows/ci.yml
```diff
@@ -30,10 +30,10 @@ jobs:
         uses: tecolicom/actions-use-apt-tools@main
         with:
           tools: linux-modules-extra-$(uname -r)
-      - name: Insert zram module
+      - name: Insert required modules
```
keszybz, line 33: OK, so this is even nicer. But please squash the first two commits together.

keszybz (changes requested):

.github/workflows/ci.yml
```diff
@@ -33,7 +33,8 @@ jobs:
       - name: Insert required modules
         run: |
           sudo depmod
-          sudo modprobe -v zram zstd
+          sudo modprobe -v zram
+          sudo modprobe -v zstd
```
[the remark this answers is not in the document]
keszybz: Oh, those three commits are touching the same lines… So I guess it's all fine if it's squashed into one commit.

PR #195, opened 2024-04-08
generator.rs: `FromIterator` is imported redundantly

pheiduck:
Until now only affects rust beta and nightly.

### DIFF
--- a/src/generator.rs
+++ b/src/generator.rs
@@ -323,6 +323,7 @@ Options={options}
 #[cfg(test)]
 mod tests {
     use super::*;
+    #[allow(unused_imports)]
     use std::iter::FromIterator;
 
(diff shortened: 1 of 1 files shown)


nabijaczleweli:
or we could just not care and not explicitly break compat

nabijaczleweli:
just mark it `#[allow(unused_imports)]`

pheiduck:
> just mark it `#[allow(unused_imports)]`

Done.

pheiduck:
I saw, that systemd is using `git rebase` so the other PR #193 was closed during that as I force push before. Will take care next time of that as well.

pheiduck:
Rust give care about that so the Patch is not needed.

Issue #178, opened 2023-04-29
Add support for new zram features (recompression, etc)


danieldjewell:
There have been changes afoot in the zram kernel module... Most notably, the addition of zram recompression: https://github.com/torvalds/linux/commit/84b33bf7888975d28c0e57011b75c445279c60ec 

see also: https://www.kernel.org/doc/html/latest/admin-guide/blockdev/zram.html#recompression

Would be nice for zram-generator to support these new features :)

nabijaczleweli:
These look like single-time triggers, so same as #164.

danieldjewell:
@nabijaczleweli As far as I'm aware: the new recompression features are automatic - no userland trigger is required, it's all handled in the kernel? (i.e. No userland daemon is required to manage/trigger anything - just like how once `zram` is setup initially and the kernel is notified to use it as a swap device, it's hands-off from there...) 

As a result, I believe it's more of a change to `zram-generator` to support the new configuration parameters?

nabijaczleweli:
The only configurable bit is `recomp_algorithm`. The `recompress` endpoint, well, recompresses the pages you tell it to, and behaves the same as `writeback` in this regard. No point configuring r_a in z-g if we don't trigger recompression v0v

polarathene:
> No point configuring `recomp_algorithm` in `zram-generator` if we don't trigger recompression v0v

Based on @nabijaczleweli response here: https://github.com/systemd/zram-generator/issues/164#issuecomment-1336507415

While `zram-generator` could generate timers based on some config to support that feature (_and I assume the same with this recompression one_), it's not being considered relevant to support and users should DIY their own?

---

Seems like the timers with basic config options would meet many users needs of those features, and it'd kind of fall into covering more zram functionality which seems relevant to the project (_simple and fast way to **configure** swap on zram_), along with it's [secondary goal of demonstrating a systemd generator in rust](https://github.com/systemd/zram-generator#rust) by additionally creating timer units? (_[probably wouldn't differ that much though..](https://github.com/systemd/zram-generator/blob/af1f29b95a3d431ca9ea3d223b398cc6e8d70201/src/generator.rs#L214-L235)_)

Only valid argument then is the convenience of having timers generated alongside the generated units, and the user config kept simple via the same `zram-generator` config?

For users that would be interested in these zram features, is there much benefit to them all DIY managing their own timers for pairing with the `zram-generator` generated units?

---

[...]

PR #200, opened 2024-09-20
Allow multiple whitespace-separated compression-algorithm=s, and give the ones after the first one to /recomp_algorithm

nabijaczleweli:
Ref: https://github.com/systemd/zram-generator/issues/178#issuecomment-2361314574

I've tested the error case, @OmegaSquad82 needs to test the "actually setting it" case.

A config like
```
compression-algorithm = a b c
```
was previously "compress with `a b c`" (naturally invalid) and is now "compress with `a`, recompress with `b` then `c`"

### DIFF
--- a/man/zram-generator.conf.md
+++ b/man/zram-generator.conf.md
@@ -72,10 +72,14 @@ Devices with the final size of *0* will be discarded.
 
   Specifies the algorithm used to compress the zram device.
 
-  This takes a literal string, representing the algorithm to use.<br />
-  Consult */sys/block/zram0/comp_algorithm* for a list of currently loaded compression algorithms, but note that additional ones may be loaded on demand.
+  This takes a whitespace-separated list string, representing the algorithms to use, and parameters in parenteses.<br />
+  Consult */sys/block/zram0/comp_algorithm* (and *.../recomp_algorithm*) for a list of currently loaded compression algorithms, but note that additional ones may be loaded on demand.
 
-  If unset, none will be configured and the kernel's default will be used.
+  If unset, none will be configured and the kernel's default will be used.<br />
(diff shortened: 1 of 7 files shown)


keszybz:
Sorry for the delay. I got the notification and then forgot to look at this.

Code-wise this looks good, but I'm not convinced that this is the right interface.
The kernel allows additional parameters to be specified, for example `type=huge_idle max_pages=42 algo=zstd`. If we allow just a list of algorithm types to be specified, it will be awkward to attach any options to this.

I wonder if the interface shouldn't be inverted, allowing the user to list each "recompression level" separately, with a list of options. But I'm not sure how to express this concisely.

nabijaczleweli:
I didn't really notice this because the interface is kinda iffy (`/recomp_algorithm` takes no options, `/recompress` takes options for the recompression subsystem *or* for a single algorithm, depending on the calling convention).

I can't think of a multi-key system that'd be any good. Maybe
```
compression-algorithm = a b(type=huge) c (threshold=3000)
```
for
```sh
echo a > /sys/block/zram0/comp_algorithm

echo algo=b    priority=1 > /sys/block/zramX/recomp_algorithm
echo type=huge priority=1 > /sys/block/zramX/recompress

echo threshold=3000       > /sys/block/zramX/recompress
```
?

keszybz:
That seems pretty nice.

keszybz:
LGTM. Let's wait for feedback from the reporter.

keszybz:
OK, let's do this. It'd be good to some testing on a real system, but we can do this later.

OmegaSquad82:
LGTM, have tested few parameters. Seems to work as expected. Thank you both.

PR #208, opened 2024-11-27
Update to latest clap and other dependencies, rework clap setup

keszybz:


nabijaczleweli:
If we're accepting having slightly-different parsing and we're not using the usage generator (what this PR ends up with is deeply hateful), can we drop clap outright?

Either way, I applied some cleanups in #209.

nabijaczleweli:
Please consider #210 in place of #208 and #209.

PR #209, opened 2024-11-27
#208 v2

nabijaczleweli:
See bits with [] for changes to original.

```
[#[rustfmt::skip] moved, attributes on expressions are experimental error]
[<String>+.clone() -> <&str>+.to_string() removes a copy]
```
```
[Drop indep]
```


keszybz:
Oh, I missed the updates here. I'll update the PR in a sec.

PR #210, opened 2024-11-27
Replace clap with getopts, bump ctor like in #208

nabijaczleweli:
The parser is equivalent to the one post-#208, except

With clap:

    $ l target/release/zram-generator
    -rwxr-xr-x 2 nabijaczleweli users 762.5k 11-27 15:40 target/release/zram-generator
    $ size target/release/zram-generator
       text    data     bss     dec     hex filename
     740420   30177     392  770989   bc3ad target/release/zram-generator

Without:

    $ size target/release/zram-generator
       text    data     bss     dec     hex filename
     577136   24457     392  601985   92f81 target/release/zram-generator
    $ l target/release/zram-generator
    -rwxr-xr-x 2 nabijaczleweli users 598.5k 11-27 16:17 target/release/zram-generator

And the dependency tree isn't infinitely large.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ exclude = ["tests/07a-mount-point-excl", "tests/10-example"]
 
 [dependencies]
 anyhow = "1.0.12"
-clap = { version = "2.33", default-features = false }
+getopts = "0.2"
 liboverdrop = "0.1.0"
 rust-ini = ">=0.15, <0.18"
 log = { version = "0.4", features = ["std"] }
--- a/src/main.rs
+++ b/src/main.rs
@@ -6,7 +6,6 @@ mod kernlog;
 mod setup;
 
 use anyhow::Result;
-use clap::{crate_description, crate_name, crate_version, App, Arg};
 use log::{info, LevelFilter};
 use std::borrow::Cow;
 use std::env;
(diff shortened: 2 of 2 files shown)


keszybz:
For example:
```console
$ clap/zram-generator --setup-device asdf asdf
error: the argument '--setup-device <device>' cannot be used with '[dir]...'

Usage: zram-generator --setup-device <device>
       zram-generator --reset-device <device>
       zram-generator dir1 [dir2 dir3]

For more information, try '--help'.

$ clap/zram-generator dir1 dir2 dir3 dir4     
error: unexpected value 'dir4' for '[dir]...' found; no more were expected

Usage: zram-generator --setup-device <device>
       zram-generator --reset-device <device>
       zram-generator dir1 [dir2 dir3]

For more information, try '--help'.

$ getopt/zram-generator --setup-device asdf asdf  
Usage:	target/debug/zram-generator dir1 [dir2 [dir3]]     # Generate systemd units
	target/debug/zram-generator --setup-device device  # Set up a single device
	target/debug/zram-generator --reset-device device  # Reset (destroy) a device

Uses /usr/lib/systemd/systemd-makefs.
```
With getopt there's always the same uninformative message.

There's also the minor detail that clap returns `2` for syntax, which can be distuinguished from `1` when operation fails.

nabijaczleweli:
We still have --help, it's just an error.

If an option-parsing error happens, it *is* logged. If a usage error happens, just the usage is logged.

I think this is all fine for something you never *ever* execute yourself.

But I'm not married to this, #209 is just as good a solution as this is imo (except for my undying hate for the direction clap keeps going).

keszybz:
For me, the dependency tree of clap is not a problem. In Fedora, clap is so widely used that it's guaranteed that it'll always be available and up-to-date.

`fedrq wrsrc rust-getopts -F source|wc -l` → 18
`fedrq wrsrc rust-getopt -F source|wc -l` → 212 
`fedrq wrsrc rust-getopt2 -F source|wc -l` → 18
`fedrq wrsrc rust-getopt3 -F source|wc -l` → 22

(The versions with number suffix are compat packages to build things that require an older version, incl. zram-generator.)

nabijaczleweli:
Your call, I don't think it really matters either way.

Issue #215, opened 2024-12-13
zram_generator::setup[…]: Warning: algorithm "…" supplemental data "…" not written: Invalid argument (os error 22)


NikitaPuzyryov:
My `/etc/systemd/zram-generator.conf` is as follows:
```
[zram0]
zram-size = 1.5 * ram
compression-algorithm = lz4 zstd(level=1) zstd(level=3) (type=idle)
swap-priority = 100
fs-type = swap
```
When restarting `dev-zram0.swap` or rebooting the machine, the following lines appear in the journal:
```
zram_generator::setup[429]: Warning: algorithm "zstd" supplemental data "algo=zstd priority=1" not written: Invalid argument (os error 22)
zram_generator::setup[429]: Warning: algorithm "zstd" supplemental data "algo=zstd priority=2" not written: Invalid argument (os error 22)
```
Am I “holding it wrong”, lack some kernel modules, or something else? I'm also not sure how to check/verify if the zram recompress machinery works how expected.

Arch Linux, kernel 6.12.4-zen1-1-zen

nabijaczleweli:
You should be seeing
```
echo lz4 > /sys/block/zram0/comp_algorithm
echo algo=zstd priority=1 > /sys/block/zram0/recomp_algorithm
echo level=1 priority=1 > /sys/block/zram0/recompress
echo algo=zstd priority=2 > /sys/block/zram0/recomp_algorithm
echo level=3 priority=1 > /sys/block/zram0/recompress
```
and then something related to the global config... which we aren't writing anywhere... for some reason (that reason being i forgor 💀 in a refactor).

You should be able to inspect the configuration by reading back those files.

And the warnings are coming from (probably!) the recompress parameter config. (You can verify this with running `zram-generator --setup-device zram0` under strace, but I'll bet that's it actually.)

Can you try #216? That should at least log correctly.

NikitaPuzyryov:
Fwiw `cat /sys/block/zram0/recomp_algorithm` shows this:
```
#1: lzo-rle lzo lz4 lz4hc [zstd] deflate 842
#2: lzo-rle lzo lz4 lz4hc [zstd] deflate 842
```
`cat /sys/block/zram0/recompress` says “Permission denied”

I'll figure out how to try your PR out when I get home. Cheers!

NikitaPuzyryov:
[...]

PR #216, opened 2024-12-13
Fix warning when configuring /recompress. Actually write global recompression configuration

nabijaczleweli:
Ref: #215


nabijaczleweli:
Closes #215

nabijaczleweli:
Rebased

XhmikosR:
Any reason why this hasn't landed yet? It seems the whole recompression feature is currently broken since I'm also hitting #215 on Raspberry Pi OS 12 (bookworm).

EDIT: although, I'm not sure I'll benefit since the Debian 12 version seems to be 1.1.2 but at least the issue will be fixed for others. Unless it's possible to update the Debian 12 upstream version after this fix has landed too.

nabijaczleweli:
Rebased. @XhmikosR you should post a debian bug and link it here, I will try to push through for updating this in a point release.

PR #223, opened 2025-02-03
config: Add warning for possibly misunderstood disk sizes

Werkov:
Kernel docs talks about expected ratio of 2:1, use that as a threshold for warning.

Link: https://lists.freedesktop.org/archives/systemd-devel/2024-November/050873.html

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -139,6 +139,17 @@ impl Device {
                 (ctx.memtotal_mb as f64 / 2.).min(4096.), // DEFAULT_ZRAM_SIZE
                 "zram-size",
             )?;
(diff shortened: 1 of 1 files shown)


nabijaczleweli:
Make it like 10 or 100 and I think this makes sense. Also, include the size in the message.

Werkov:
Thanks for the review, I made the message more explicit and changed the threshold to logarithmic midpoint of your suggestions ;-)

FTR, I still think it's super confusing that 800M results in 800e6 MiB. I hope the warning helps when backwards incompatible changes are discouraged.

PR #224, opened 2025-02-23
Cargo.toml: bump to edition 2024

pheiduck:


nabijaczleweli:
Why? this doesn't do anything except exclude building with compilers older than a week and maybe change Some behaviour that had worked into Some new behaviour that doesn't. The point of editions is that you can write your project against a certain set of semantics without then them magically exploding. I can't say I see a reason to ever bump the edition.

Issue #205, opened 2024-11-21
zram service not starting automatically after reboot


PamAllay:
I have followed arch wiki on configuring zram using zram-generator:
https://wiki.archlinux.org/title/Zram

but the systemd-zram-setup@zram0.service service does not start on boot as stated in the wiki, i manually have to run : sudo systemctl start systemd-zram-setup@zram0.service to start the service after every boot, and i've had no luck finding a solution.

configuration = 
$>cat /etc/systemd/zram-generator.conf 
[zram0]
zram-size = min(ram / 2, 4096)
compression-algorithm = zstd

keszybz:
The generator makes `dev-zram0.swap` part of `swap.target`, and `dev-zram0.swap` pulls in `systemd-zram-setup@zram0.service`. If that doesn't happen for you, then it sounds like for some reason `swap.target` is not getting activated properly.

```console
$ systemctl list-dependencies sysinit.target
sysinit.target
...
● ├─swap.target
● │ └─dev-zram0.swap
...
```

PamAllay:
@keszybz i got this ->
○ ├─swap.target
○ │ ├─dev-disk-by\x2duuid-708fd83a\x2d61fd\x2d4413\x2d9c05\x2da5d56ea0ea93.swap
● │ └─dev-zram0.swap

the first swap must be the swap on disk and is disabled by kernal parameter and zram swap is enabled as it has highlighted dot, any help..?

keszybz:
`swap.target` is also not active (white dot). This seems to confirm my original theory that `swap.target` is not getting activated. Can you show the kernel commandline?

PamAllay:
that is also a static type, so can't enable it.. any way to activate it?

PR #233, opened 2025-05-31
Adding the ability to let zram-generator service autostart.

liamthexpl0rer:
Hi there,

I added the ability to use `systemctl enable systemd-zram-setup@zramN.service`.

This is my first pull request at all so please don't hesitate to tell me if I'm doing it right. ;-)

### DIFF
--- a/units/systemd-zram-setup@.service.in
+++ b/units/systemd-zram-setup@.service.in
@@ -13,3 +13,6 @@ Type=oneshot
 RemainAfterExit=yes
 ExecStart=@SYSTEMD_SYSTEM_GENERATOR_DIR@/zram-generator --setup-device '%i'
 ExecStop=@SYSTEMD_SYSTEM_GENERATOR_DIR@/zram-generator --reset-device '%i'
+
+[Install]
+WantedBy=multi-user.target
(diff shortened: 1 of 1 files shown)


liamthexpl0rer:
Should also fixing #205 .

nabijaczleweli:
This is wrong: (a) this is WantedBy swap.target, and (b) the generator *already* does this:
```rust
    /* enablement symlink */
    let symlink_path = output_directory.join("swap.target.wants").join(&swap_name);
    let target_path = format!("../{swap_name}");
    make_symlink(&target_path, &symlink_path)?;
```

nabijaczleweli:
This is *already* done by the generator. Do you have a real use-case that's affected/fixed by this?

liamthexpl0rer:
@nabijaczleweli 
a) I have changed it to `WantedBy=swap.target`
b) It doesn't work. When I use `systemctl start systemd-zram-setup@zram0.service` after a reboot zram0 is not active anymore. On `systemctl enable systemd-zram-setup@zram0.service` i become the error, that systemd is missing the installation config.

nabijaczleweli:
The dependency chain is sysinit.target wants swap.target wants wants dev-zram0.swap requires systemd-zram-setup@zram0.service. Just starting systemd-zram-setup@zram0.service won't make the swap become real, and if it does then there's an issue somewhere else in the dependency chain.

liamthexpl0rer:
@nabijaczleweli Okay, but enabling `systemd-zram-setup@zram0.service` works for me.
Can you tell me, what is the right way to have a permanent swap? I have already searched on the internet for a manual, but I didn't found it. Your readme is also not really helpful for solving my problem.

nabijaczleweli:
If, for some reason, the initialisation doesn't get triggered, then `systemctl start swap.target` will trigger it. *That reason* will be the issue, and you've already identified it as reported in #205.

PR #237, opened 2025-10-14
setup: send supplemental params to algorithm_params

Szpadel:
zram-generator used to route secondary compressor parameter blobs to /sys/block/zramX/recompress while setting up the device. The kernel's recompress knob refuses writes until the zram device is initialised (disksize set), so CONFIG_ZRAM_MULTI_COMP setups emitted EINVAL warnings like:

  Warning: algorithm "zstd" supplemental data "algo=zstd priority=1" not written: Invalid argument

Route all supplemental data through /sys/block/zramX/algorithm_params instead, tagging secondary algorithms with their priority. This matches the driver interface, prevents the warning, and still supports primary and secondary parameter payloads. Include unit coverage for the new helper that builds the payload.


analytically:
I'd also like to see this merged, thank you

PR #240, opened 2026-02-26
Fix error in `zram-generator.conf.example`

Atrate:
At least on my system, `xenstore-read` always returns values in `KiB`, not `MiB`. This lead to an error in the example config. This PR also simplifies the `xenstore-read` call with a shorthand, thanks to @marmarek for that idea.


Atrate:
Additionally, it may be a good idea to change `hotplug-max` to `static-max`, as HVMs only have the latter while PVHs have both.

marmarek:
> Additionally, it may be a good idea to change `hotplug-max` to `static-max`, as HVMs only have the latter while PVHs have both.

But those are not equivalent. `static-max` is a max value you can reach when using balloon driver, and `hotplug-max` is max you can reach with memory hotplug. If memory hotplug is enabled, usually it's used instead of balloon, so `static-max` just means the current allocation (or max usage since startup) - not the actual max value.

What you see is Qubes OS defaults: PVH uses hotplug, but HVM does not. And also, most use of HVM (at least by default) is for PCI devices, which also makes it not really compatible with any kind of dynamic memory management, so in that case you can as well take the size Linux sees at boot, not some Xen-specific value, as it won't change at runtime.

Atrate:
As far as I'm reading this, then, there will not be one solution for each type of Qubes/Xen VM. Let's leave the example as-is, this got too Xen-specific for this repo.

